PR #27, opened 2024-05-08
Logging

briri:
- Added Pino logger and the ECS formatter.
- Added an Apollo Server plugin that uses the logger to record Apollo Server request/response lifecycle events
- Added the logger to the Apollo Server context and updated the resolvers to use it

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,8 +4,10 @@ import { addMocksToSchema } from '@graphql-tools/mock';
 import { makeExecutableSchema } from '@graphql-tools/schema';
 
 import { mysqlConfig } from './config';
+import { logger } from './logger';
 import { MysqlDataSource } from './datasources/mysqlDB';
 import { DMPHubAPI } from './datasources/dmphub-api';
+import { loggerPlugin } from './plugins/logger';
 
 import { typeDefs } from './schema';
 import { resolvers } from './resolver';
@@ -32,6 +34,9 @@ function serverConfig() {
 
 async function startApolloServer() {
   const apolloConfig = { ...serverConfig(), ...{
+    // The LoggerPlugin is used by Apollo server to record events in the request/response
+    // lifecycle as well as handling any GraphQL errors
+    plugins: [loggerPlugin(logger)],
     // Mitigation for an issue that causes Apollo server v4 to return a 200 when a query
     // includes invalid variables.
     //    See: https://www.apollographql.com/docs/apollo-server/migration/#known-regressions
@@ -45,6 +50,8 @@ async function startApolloServer() {
       const token = getTokenFromRequest(req);
 
       return {
+        // Pass the logger in so it is available to our resolvers and dataSources
+        logger,
         dataSources: {
           dmphubAPIDataSource: await new DMPHubAPI({ cache, token }),
           sqlDataSource: await new MysqlDataSource({ config: mysqlConfig }),
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,0 +1,22 @@
+import pino, { Logger } from 'pino';
+import { ecsFormat } from '@elastic/ecs-pino-format';
+
+const logLevel = process.env.LOG_LEVEL || 'info';
+
+/*
+ * Logger using Pino and the Elastic Common Schema format to facilitate
+ * standardization across all log messaging. This format will allow us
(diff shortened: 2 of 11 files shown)


src/logger.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ecsFormat } from '@elastic/ecs-pino-format';
+
+const logLevel = process.env.LOG_LEVEL || 'info';
+
+/*
+ * Logger using Pino and the Elastic Common Schema format to facilitate
+ * standardization across all log messaging. This format will allow us
+ * to more easily debug and track requests in OpenSearch.
+ */
+export const logger: Logger = pino({ level: logLevel,  ...ecsFormat })
```
jupiter007, line 11: @briri Would we ever want a dynamic way to update the level in production, if we were trying to debug a production issue?

src/resolvers/dmsp.ts
```diff
@@ -1,13 +1,26 @@
...
       return new Promise((resolve, reject) => {
         dataSources.dmphubAPIDataSource.getDMSP(encodeURIComponent(dmspId))
-          .then(rows => resolve(rows))
-          .catch(error => reject(error));
+          .then(rows => {
+            formatLogMessage(logger).debug(logMessage);
+            resolve(rows)
+          })
+          .catch(err => {
+            formatLogMessage(logger, { err, dmspId }).debug(`ERROR: ${logMessage} - ${err.message}`);
```
jupiter007, line 21: @briri  - Line 21 - If you are logging this as an "ERROR", would it make more sense to use ".error()" instead of ".debug()" on this?

jupiter007 (commented):
@briri - This looks good to me. I just had a few comments/questions.

src/logger.ts
jupiter007: Ok, thanks @briri. It is probably overkill. It was just a thought because I came across a situation before where I needed to change logging levels in production.

src/logger.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// Add the additional args to the ECS logger so that they are queryable fields.
+// At least you should send the sessionId if available and the error (if applicable)
+export function formatLogMessage(logger: Logger, args: Object | void): Logger {
+  if (args) {
+    // Add the args to the ECS logger
+    return logger.child({ ...args });
+  }
+  // Otherwise there were no additional arfs, so return the logger as-is
```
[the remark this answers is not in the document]
briri: haha, arts 🐶, good catch!

src/plugins/logger.ts
```diff
@@ -0,0 +1,133 @@
...
+  return loggerInstance.child({
+    httpMethod: request.http?.method,
+    httpReferer: hdrs?.get('referer'),
+    httpSecFetchSite: hdrs?.get('sec-fetch-site'),
+    httpUserAgent: hdrs?.get('user-agent'),
+    // TODO: We will eventually want to update this to include the nextJS session id
+    operationName: request.operationName,
+    variables: request.variables,
+    query: request.query,
+    err: errs
```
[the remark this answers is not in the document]
briri: yes! and would be super useful information. I will add it

src/resolvers/dmsp.ts
briri: yep. 💯

src/index.ts
```diff
@@ -1,24 +1,97 @@
...
+          })
+          .catch((error) => {
+            logger.error(`ERROR: Healthcheck - ${error.message}`);
+            res.status(400).send(JSON.stringify(error));
+          });
+  });
+
+  // Set up our Express middleware to handle CORS, body parsing, and our expressMiddleware function.
+  app.use(
+    '/',
```
jupiter007, line 64: @briri - eventually, when "app.use("/")" gets sort of larger, you could move out middleware functions into a separate locations, like 
"/middleware", and maybe just call them using something like:

```
app.use("/", (req, res, next) => {
   handleCors(req,res);
   extractToken(req, res);
   next();
});
```
or something like that.

briri: Yes, great idea. Better to do it now too while the code is relatively small. I moved the middleware into its own files and also moved the health check code into a separate file since we will be defining a few other non-graphQL pages to handle auth.

jupiter007: @briri  Nice!

PR #29, opened 2024-05-31
Mock store

briri:
Added a custom MockStore that can be used in our docker dev environment.

The temporary store is ephemeral and resets when the application restarts. 

There are also some changes in the ./data-migrations folder that aren't related to the MockStore. They were added to allow me to seed our database for the new ECS containers

### DIFF
--- a/src/mocks/contributorRole.ts
+++ b/src/mocks/contributorRole.ts
@@ -1,37 +1,68 @@
+import casual from 'casual';
+import { mockDate, mockError, mockSuccess } from '../mocks';
+import {
+  ContributorRole,
+  QueryContributorRoleByIdArgs,
+  QueryContributorRoleByUrlArgs
+} from '../types';
+
 // Just define what is unique here. Any fields you skip will end up using the defaults
 // for their respective type as defined in ./src/mocks.ts
-const data = [
+const data: ContributorRole[] = [
   {
-    id: 1,
+    id: '1',
+    displayOrder: 1,
     url: 'https://credit.niso.org/contributor-roles/investigation/',
     label: 'Principal Investigator (PI)',
     description: 'An individual conducting a research and investigation process, specifically performing the experiments, or data/evidence collection.',
-    displayOrder: 1,
+    created: mockDate,
+    modified: mockDate
   },
   {
-    id: 2,
+    id: '2',
     url: 'https://credit.niso.org/contributor-roles/project-administration/',
     label: 'Project Administrator',
     description: 'An individual with management and coordination responsibility for the research activity planning and execution.',
     displayOrder: 2,
+    created: mockDate,
+    modified: mockDate
   },
   {
-    id: 3,
+    id: '3',
     url: 'https://credit.niso.org/contributor-roles/data-curation/',
     label: 'Data Manager',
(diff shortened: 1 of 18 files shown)


src/mocks/contributorRole.ts
```diff
@@ -1,37 +1,105 @@
...
-const data = [
+import { Resolvers } from "../types";
+import { MockMySQLTable } from './MockMySQLTable';
+// import { ContributorRole, ContributorRoleMutationResponse } from '../types';
+
+// Seed records for the ContributorRoles table
+const mockItems = [
   {
-    id: 1,
+    displayOrder: 1,
```
jupiter007, line 8: L8: The other objects in the array  have "displayOrder" property last. Should we keep them consistent?

src/mocks/contributorRole.ts
```diff
@@ -1,37 +1,105 @@
...
+export const mockStore = new MockMySQLTable(mockItems);
+
+export const resolvers: Resolvers = {
+  Query: {
+    // returns an array of all contributor roles
+    contributorRoles: (_, __, { mockStores }) => {
+      return mockStores.contributorRoles.items();
+    },
+
+    // No idea why this does not work without including `_other` here. Results in a Typescript error
```
jupiter007, line 41: L41: Does this comment still apply? I don't I understand the "_other" comment.

jupiter007:
For index.ts, I wonder if we could change the handling of the async to be like the following. It was easier for me to read using `async-await`.
 
```
import { ApolloServer } from '@apollo/server';

import express from 'express';
import http from 'http';

import { logger } from './logger';
import { serverConfig } from './config';
import { healthcheck } from './pages/healthcheck';
import { handleCors } from './middleware/cors';
import { attachApolloServer } from './middleware/express';

// Required logic for integrating with Express
const app = express();
// Our httpServer handles incoming requests to our Express app.
const httpServer = http.createServer(app);


const server = new ApolloServer(serverConfig(logger, httpServer));

const startServer = async () => {
    await server.start();
    const { cache } = server;

    // Healthcheck endpoint (declare this BEFORE CORS definition due to AWS ALB limitations)
    app.get('/up', (_request, response) => healthcheck(server, response, logger));

    // Express middleware
    app.use(
        '/',
        // 50mb is the limit that Apollo `startStandaloneServer` uses.
        express.json({ limit: '50mb' }),
        // CORS config
        handleCors(),
        // Attach Apollo server
        attachApolloServer(server, cache, logger),
    );
    // TODO: Add our auth and token endpoints here

    httpServer.listen({ port: 4000 }, () => {
        console.log(`Server ready at http://localhost:4000/graphql`);
    })
}


startServer().catch((error) => {
    console.log('Error starting server:', error)
})

```

jupiter007:
Also, @briri if you have time on Monday, could you help me figure out why I can't run the data migrations locally? I wanted to try running the app with the mocks, but I wasn't able to get the data.

jupiter007:
I'm wondering, in logger.ts, whether you should add another "optional chaining" on line 17, so instead of "`httpMethod: context?.request.http?.method`,", it is "`httpMethod: context?.request?.http?.method`".

briri:
Closing this one since we got the `preserveResolver` working as expected

Issue #31, opened 2024-06-04
Set up some basic authentication handling with JWT


jupiter007:
We want to set up a simple "username, password" authentication initially with JWT. This will allow us to test some basic authentication and to set up some infrastructure for it.

jupiter007:
I added some initial code to help us run some initial tests and think about infrastructure:

https://github.com/CDLUC3/dmsp_backend_prototype/tree/feature/simple-authentication

PR #32, opened 2024-06-05
Implemented a simple username, password authentication

jupiter007:
[#31](https://github.com/CDLUC3/dmsp_backend_prototype/issues/31)

Created a router.js file to handle the different routes, and added a User controller to handle the different auth endpoints, as well as a User model/class.

Used 'jsonwebtoken' to generate a token for the frontend and 'bcrypt' to encrypt the password for the database.

Also, added a couple of .sql files to the data-migration scripts to create a User table, and populate it with a test user.


jupiter007:
I went the changes with Brian, and we agreed to go ahead and merge the changes into the development branch.

PR #34, opened 2024-06-07
Feature/combined auth

briri:
@jupiter007 I am not 100% done with this yet, but I wanted to put the code changes out here to get your feedback. There is still some cleanup and a lot of testing to do.

- I added OAuth2 
- I also refactored your code a bit, @jupiter007  (mostly just moving things around). 

I have not tested any of it. I finally got all of the Typescript issues resolved and wrapped my head around the `@node/express-oauth-server` documentation. I am sure there will be some issues that I need to iron out.

Both the JWT auth you setup and the OAuth2 access token use the same `generateToken` function. I am hoping that this will be enough to allow the GraphQL endpoints to work with a single token and not care where the auth came from.

I plan to test it out on Monday and start writing some Jest tests


.DS_Store
jupiter007, line 1: Please remember to add this file to .gitignore

jupiter007:
Hi @briri. I added a few comments to your PR. There's a lot for my brain to process, but I think it's looking good so far.

I noticed that you took the "feature/simple-authentication" branch to work off of, but I thought I merged that branch into development. I think you should make sure to merge the latest development branch into this one to make sure you have all the changes I made to fix the pipeline after my merge.

src/middleware/oauthServer.ts
```diff
@@ -0,0 +1,115 @@
...
+    } catch(err) {
+      console.log(err);
+      return null;
+    }
+  },
+
+  // Verify that the scope requested is allowed by the access token
+  verifyScope: async function(token: Token, scope: string | string[]) {
+    const scopeArray: string[] = Array.isArray(scope) ? scope : [scope];
+    return await scopeArray.every((item) => token.scope.includes(item));
```
jupiter007, line 47: L47: This looks like a synchronous function. I don't think you need the "`await`"

src/router.ts
```diff
@@ -1,25 +1,40 @@
...
+// Sample protected endpoint
 router.post("/protected", authMiddleware, (req, res) => {
-    if (!req.auth.admin) {
-        return res.sendStatus(401);
-        res.sendStatus(200);
-    }
+  // TODO: Someday fix this so we aren't using `any`
+  if (!(req as any).auth.admin) {
+    return res.sendStatus(401);
+    res.sendStatus(200);
```
jupiter007, line 36: L36 - `res.sendStatus(200)` should be moved to after L37.

src/router.ts
```diff
@@ -1,25 +1,40 @@
...
+  credentialsRequired: false,
+  secret: generalConfig.jwtSecret as string,
 });
 
+// Support for email+password
+router.post('/signin', (req, res) => signinController(req, res));
+router.post('/signup', (req, res) => signupController(req, res));
+
+// Support for OAuth2
+router.get('/authorize', (req, res) => oauthServer.authorize(castRequest(req), castResponse(res)));
```
jupiter007, line 24: One thing I'm not clear about is how the frontend will get the Client ID. Does it request it from the backend server?

src/models/OAuthClient.ts
```diff
@@ -0,0 +1,148 @@
...
+  generateClientSecret(): string {
+    return uuidRandom();
+  }
+
+  // Register/Save a new OAuthClient
+  async save(): Promise<boolean> {
+    this.cleanup();
+
+    let sql = `
+      INSERT INTO aouthClients (name, redirectUris, grants, clientId, clientSecret, userId)
```
jupiter007, line 92: L92: Typo '`aouthClients`' should be `oauthClients`

src/models/OAuthClient.ts
```diff
@@ -0,0 +1,148 @@
...
+      this.clientSecret,
+      this.user?.id?.toString(),
+    ];
+
+    // If the OAuthClient already has an id then update it instead
+    if (this.id) {
+      vals.push(this.id);
+      vals.push(new Date().toISOString());
+      sql = `
+        UPDATE aouthClients
```
jupiter007, line 109: L109: Typo `aouthClients` should be `oauthClients`

src/models/OAuthCode.ts
```diff
@@ -0,0 +1,174 @@
...
+      console.log('Error revoking OAuthCode: ', err)
+      throw err;
+    }
+  }
+
+  // Convert the DB fields to properties of this call (except client and user)
+  static _SqlFieldsToProperties(row) {
+    const scopeArray = typeof row.scope === 'string' ? row.scope.split(' ') : row.scope;
+    return {
+      authorzationCode: row.code,
```
jupiter007, line 166: L166: Typo - `authorzationCode` should be `authorizationCode`

briri: good catch

src/helpers.ts
```diff
@@ -0,0 +1,14 @@
+
+// Convert a string into an Array (return the default or an empty array if it is null or undefined)
+export function stringToArray(array: any, dflt: string[] = []): string[] {
```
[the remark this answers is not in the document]
briri: It didn't like 'default' it said it was a reserved word so I will switch to be defaultResponse.

Thanks. I had been wondering between the difference between `typeof` and `instanceof`. Your explanation makes sense

.env-example
```diff
@@ -10,6 +10,10 @@ USE_MOCK_DATA=false
 # and then provide fill out the following variables
 AWS_REGION=us-west-2
 
+# Salt used to hash DB fields
+SALT=$2a$10$HeRb8Z2NK86pjJp300/flO
```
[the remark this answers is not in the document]
briri: yes. We will add these values to SSM for the AWS environment

src/models/User.ts
```diff
@@ -0,0 +1,127 @@
...
+    this.role = this.role || UserRole.Researcher;
+    this.givenName = this.givenName?.trim();
+    this.surName = this.surName?.trim();
+  }
+
+  // Verify that the email does not already exist and that the required fields have values
+  async validate(): Promise<boolean> {
+    // check if email is already taken
+    const existing = await User.findByEmail(this.email);
+    if (existing) {
```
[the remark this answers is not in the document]
briri: Yes, great point.

src/models/User.ts
```diff
@@ -0,0 +1,127 @@
...
+      console.error(`Error logging in User: ${this.email}`);
+      throw err;
+    }
+  }
+
+  // Register the User if the data is valid
+  async register(): Promise<boolean> {
+    this.cleanup();
+    let salt = bcrypt.genSaltSync(10);
+    this.password = bcrypt.hashSync(this.password, generalConfig.salt);
```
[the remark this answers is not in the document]
briri: I was confused by that too. I'm not sure. Will ping you on slack to discuss

src/router.ts
briri: It won't need to use this route. It will use the sign in/up routes

src/services/tokenService.ts
```diff
@@ -0,0 +1,17 @@
+import jwt from 'jsonwebtoken';
+import { User } from '../models/User';
+import { generalConfig } from '../config/generalConfig';
+
+// Generate a JWT Token for the given User
+export const generateToken = (user: User): string => {
+  return jwt.sign(
+    { id: user.id, email: user.email, role: user.role.toString },
```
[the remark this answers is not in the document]
briri: The sign in should be fetching that info from the DB when we call `user.login()`. 

I will update the `user.register()` to return the user instead of a boolean so that its available when the token is generated

src/router.ts
jupiter007: I did some reading about the Client ID and Client Secret yesterday, and it seems that the frontend would get it from the OAuth provider and store it somewhere safe.

src/services/tokenService.ts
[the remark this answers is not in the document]
jupiter007: Ok thanks.

briri (commented):
Thanks for adding the tests here @jupiter007! I can use this as a template for my other tests :)

jupiter007:
Nice linting and tests

jupiter007 (approved):

PR #40, opened 2024-07-02
Fix an issue with logger reference that was breaking build

briri:
- fix issue with `logger` causing a build failure
- removed debug `console.log` statements
- updated other remaining `console.log` lines to use our formatted logger

### DIFF
--- a/src/resolvers/__tests__/user.spec.ts
+++ b/src/resolvers/__tests__/user.spec.ts
@@ -37,6 +37,7 @@ describe('User queries', () => {
     expect(res.body.singleResult.data?.users).toBeDefined();
 
     const result = res.body.singleResult.data?.users as [User];
+    // console.log(result);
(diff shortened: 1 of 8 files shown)


src/resolvers/__tests__/user.spec.ts
```diff
@@ -37,6 +37,7 @@ describe('User queries', () => {
     expect(res.body.singleResult.data?.users).toBeDefined();
 
     const result = res.body.singleResult.data?.users as [User];
+    // console.log(result);
```
jupiter007, line 40: L40: Did you mean to leave this commented out "console.log()" line?

jupiter007 (approved):

src/resolvers/__tests__/user.spec.ts
briri: yeah, that test is currently failing so I need to get back to it

PR #42, opened 2024-07-06
Small updates to authentication code after testing with frontend

jupiter007:
- I added a check in signinController and signupController for token, because "generateToken" could return null
- Update signupController to use the pipe delimiter ('|') for creating a string of concatenated errors, because commas are used in error messages, and it makes it difficult for frontend to create an array from the string.
- Updated User.spec.ts so that tests pass
- Also, made some comments on specific changes below for more details

### DIFF
--- a/src/controllers/signinController.ts
+++ b/src/controllers/signinController.ts
@@ -10,7 +10,12 @@ export const signinController = async (req: Request, res: Response) => {
 
     if (user) {
       const token = generateToken(user);
-      res.json({ success: true, token });
+      if (token) {
+        res.status(200).json({ success: true, token });
+      } else {
+        throw new Error('Login failed');
+      }
+
     } else {
       res.status(401).json({ success: false, message: 'Invalid credentials' });
     }
--- a/src/models/User.ts
+++ b/src/models/User.ts
@@ -135,7 +135,7 @@ export class User {
     const email = this.email || '';
 
     if (!validateEmail(email) || !this.validatePassword()) {
-      this.errors.push('Login failed');
+      return null;
     }
 
     try {
@@ -154,7 +154,7 @@ export class User {
   // Register the User if the data is valid
   async register(): Promise<User | Falsey> {
     this.cleanup();
-    this.validateNewUser();
+    await this.validateNewUser();
 
     if (this.errors.length === 0) {
       const passwordHash = await this.hashPassword(this.password);
@@ -175,7 +175,7 @@ export class User {
       }
     } else {
       formatLogMessage(logger)?.debug(`Invalid user: ${this.email}`);
-      return null;
+      return this;
     }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,3 +18,6 @@
 - Pino logger with ECS formatter
 - Plugin to log request/response lifecycle events
 - Add Logger to the context and then used it in the resolvers
+
+### Updated
+- Made some updates to auth code based on testing out recent changes with frontend [#34]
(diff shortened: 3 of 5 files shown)


src/models/User.ts
```diff
@@ -135,7 +135,7 @@ export class User {
     const email = this.email || '';
 
     if (!validateEmail(email) || !this.validatePassword()) {
-      this.errors.push('Login failed');
+      return null;
```
jupiter007, line 138: Returned null, because the signinController that is calling this method is not checking for user.errors, but just wants to know if the response is null or not.

src/models/User.ts
```diff
@@ -154,7 +154,7 @@ export class User {
   // Register the User if the data is valid
   async register(): Promise<User | Falsey> {
     this.cleanup();
-    this.validateNewUser();
+    await this.validateNewUser();
```
jupiter007, line 157: Since validateNewUser is async we needed to add "`await`"

src/models/User.ts
```diff
@@ -175,7 +175,7 @@ export class User {
       }
     } else {
       formatLogMessage(logger)?.debug(`Invalid user: ${this.email}`);
-      return null;
+      return this;
```
jupiter007, line 178: We need to return the user object, because the signupController is expecting the user object with its errors, so that this info can be passed to the frontend.

src/controllers/signinController.ts
```diff
@@ -10,7 +10,12 @@ export const signinController = async (req: Request, res: Response) => {
 
     if (user) {
       const token = generateToken(user);
-      res.json({ success: true, token });
+      if (token) {
+        res.status(200).json({ success: true, token });
+      } else {
+        throw new Error('Login failed');
```
briri, line 16: should these respond by adding a status code and message to the response object instead? Not sure what Express' error handling looks like for this

briri (changes requested):
I'm fine with whichever makes sense. 
Does Express have specific handling if you throw an Error in the controller. For example I've seen it will pass along to `next()` when it finishes one thing, so does that stop it immediately?
Whichever one we choose though, we should make them all the same

src/controllers/signinController.ts
jupiter007: Good question @briri . When you throw an error here, it will be caught by the "catch" in the "try-catch" that it's in. This is what will be logged: `"{"level":50,"time":1720547549009,"pid":42151,"hostname":"CDL-JSHIN-9M","err":{"type":"Error","message":"Login failed","stack":"Error: Login failed\n    at signinController (/Users/jshin/dev/dmsp_backend_prototype/src/controllers/signinController.ts:16:15)"},"msg":"Signin error"}"` and it will return a 500 error with the message "Internal Server Error".

jupiter007:
> I'm fine with whichever makes sense. Does Express have specific handling if you throw an Error in the controller. For example I've seen it will pass along to `next()` when it finishes one thing, so does that stop it immediately? Whichever one we choose though, we should make them all the same

Yes, in NodeJS Express, the middleware has a "next()" function in the request-response cycle. You typically want to have an error handler as one of the last middlewares to catch any uncaught errors. We should probably add at least one catch-all error handler middleware.

briri (approved):

jupiter007:
> I'm fine with whichever makes sense. Does Express have specific handling if you throw an Error in the controller. For example I've seen it will pass along to `next()` when it finishes one thing, so does that stop it immediately? Whichever one we choose though, we should make them all the same

Hi @briri . Are you ok with resolving this comment so I can merge?

PR #70, opened 2024-08-22
Updates to fix AWS deploy issues

### LINKED ISSUE #71: Issues deploying the updated applications to AWS environment
Discovered multiple issues when trying to deploy the updated backend code to the AWS environment.

1. The backend modified the login and registrations endpoints to be `/signin` and `signup`. This conflicted with the `/signin` endpoint on the frontend. It worked fine when developing locally, but the servers reside at the same domain name and we are using ALB (load balancer) rules to direct traffic between the nextJS and Apollo Server based on the path.
2. The backend was missing the JWT_SECRET environment variables and several others
3. I was unable to run the data-migration scripts on the Cloud9 Bastion instance. This was due to it not using a DotEnv file but instead wanting to use the AWS CLI to fetch these values from SSM and CF stack outputs

I also discovered that the frontend URLs were `https://ui.dmphub.uc3dev.cdlib.net/undefined/[backend path]`. I encountered this when running locally after the changes to the ENV variables for that app.

briri:
## Description

We ran into some issues deploying the latest development branch to AWS.

Fixes #71

## Type of change
- [x] Bug fix (potentially breaking change which fixes an issue)

The data-migrations can now be run locally or in the AWS Cloud9 bastion instance.

The new routes "may" break the connectivity between the frontend and the backend. @jupiter007 can you check the frontend code to see if that is true? I think it will depend on whether the code takes in the URL from the env variable and adds a `/` between it and the path (e.g. `signing`). 
- If it does not, then we can just update the CloudFormation template to include the `/apollo-` prefix (e.g. `https://ui.dmphub.uc3dev.cdlib.net/apollo-`) 
- if it does, then the frontend will need to be updated to include the `apollo-` prefix

## How Has This Been Tested?
The data-migrations were run manually and successful.

Cannot test the new router endpoints until the app is deployed. Will need the associated AWS repo PR deployed.


src/router.ts
```diff
@@ -14,13 +14,13 @@ const authMiddleware = expressjwt({
 });
 
 // Support for email+password
-router.post('/signin', (req, res) => signinController(req, res));
-router.post('/signup', (req, res) => signupController(req, res));
+router.post('/apollo-signin', (req, res) => signinController(req, res));
```
briri, line 17: @jupiter007 you may need to update these in the frontend

data-migrations/nuke-db.sh
```diff
@@ -1,11 +1,31 @@
-# Load the Dotenv file
-. .env
 
-# Run the migration using the env variable defined in the dotenv file
-if [ -z "${MYSQL_PASSWORD}" ]; then
-  MYSQL_ARGS="-h ${MYSQL_HOST} -P ${MYSQL_PORT} -u ${MYSQL_USER}"
+if [ ! -f /tmp/foo.txt ]; then
```
jupiter007, line 2: @briri  is "/tmp/foo.txt" correct?

data-migrations/process.sh
```diff
@@ -1,15 +1,28 @@
...
 
-# =============================================================
-# = This script is meant to be run in the AWS CodeBuild step! =
-# =============================================================
-. .env
-
-# Run the migration using the env variable defined in the dotenv file
-if [ -z "${MYSQL_PASSWORD}" ]; then
-  MYSQL_ARGS="-h ${MYSQL_HOST} -P ${MYSQL_PORT} -u ${MYSQL_USER}"
+if [ ! -f /tmp/foo.txt ]; then
```
jupiter007, line 3: L3 - @briri - same with this. What is "tmp/foo.txt" for?

src/router.ts
jupiter007: Thanks @briri . I created this PR to go with this change: https://github.com/CDLUC3/dmsp_frontend_prototype/pull/100

jupiter007 (approved):

data-migrations/nuke-db.sh
briri: lol. copy/paste.

Will update these and the changelog!

PR #83, opened 2024-09-09
Feature/add section migrations

### LINKED ISSUE #74: Build DB Migrations for Sections
Split task from https://github.com/CDLUC3/dmsp_backend_prototype/issues/63

Requires Schema first #73 

Copy real ones from existing DMP Tool

briri:
## Description

- Data migrations to create MySQL tables for `sections`, `versionedSections`, `tags` and `sectionTags`
- Data migrations to seed the new tables (used output from the [a DMPTool repo branch](https://github.com/CDLUC3/dmptool/tree/chore/data-migration) to use real template and section data)
- Moved the SQL for the data migration that creates the `dataMigrations` table into the `database-init.sh` and `nuke-db.sh` scripts
- Updated the `seed-users` migration to include a mock admin user for each org that has a template
- Removed the old `seed-templates` migration 

Fixes #74

## Type of change

## How Has This Been Tested?

1. Nuke your local DB and rebuild using the scripts `data-migrations/nuke-db.sh && data-migrations/process.sh`.
2. Then reviewed data in the new tables via Sequel Pro

## Checklist:


data-migrations/2024-09-05-0945-seed-tags.sql
```diff
@@ -0,0 +1,32 @@
+INSERT INTO tags (name, description, createdById, modifiedById)
```
jupiter007, line 1: @briri - where do these tags come from? They are different than the Themes, and I couldn't remember if we had another document listing these out somewhere.

data-migrations/database-init.sh
```diff
@@ -27,8 +27,18 @@ CREATE_DATABASE="CREATE DATABASE IF NOT EXISTS ${MYSQL_DATABASE}
   CHARACTER SET utf8mb4
   COLLATE utf8mb4_unicode_ci;"
 
-echo "-h${MYSQL_HOST} -P${MYSQL_PORT} -u${MYSQL_USER} -p ${MYSQL_PASSWORD}"
-
 # Create the Database
 echo "Creating database ${MYSQL_DATABASE} ..."
-mysql -h${MYSQL_HOST} -P${MYSQL_PORT} -u${MYSQL_USER} -p ${MYSQL_PASSWORD} <<< ${CREATE_DATABASE}
+mariadb -h${MYSQL_HOST} -P${MYSQL_PORT} -u${MYSQL_USER} -p${MYSQL_PASSWORD} <<< ${CREATE_DATABASE}
```
jupiter007, line 32: @briri - Should this file be mentioned in the README.md file to initialize the database?

data-migrations/2024-09-05-0945-seed-tags.sql
jupiter007: Thanks for that info @briri.

jupiter007:
Looks good to me @briri.

jupiter007 (approved):

jupiter007:
@briri , I did notice that super@example.com user is the only one with data in the `templates` table. Do you think we should seed any data for the other initial test users.

jupiter007:
> Ah I see. Well, we could I suppose connect them to one of the other org's templates, or create samples for them. Maybe it's good to have some users with no templates though so we can see what pages look like when the user is new or hasn't created any templates. Most orgs do not have their own templates, they just customize funder templates with additional guidance.
> 
> If you login as one of the new admins like `admin@unicamp.br` you should see templates.

Ok. That makes sense. Thanks Brian.

Issue #75, opened 2024-08-28
Build Models + Resolvers for Sections


bofstein:
Split task from https://github.com/CDLUC3/dmsp_backend_prototype/issues/63

Requires these first: #73 #74

Issue #73, opened 2024-08-28
Build Schema for Sections


bofstein:
Split task from https://github.com/CDLUC3/dmsp_backend_prototype/issues/63

jupiter007:
@bofstein and @briri, I think I remember something about not including Tags in the Section page because it was confusing people. Can you confirm that we need Tags and/or SectionTags?

PR #89, opened 2024-09-17
Added resolvers and models for Section, Tag, and VersionedSection

jupiter007:
## Description

After the schemas were added for Section, Tag and VersionedSection in ticket [#73](https://github.com/CDLUC3/dmsp_backend_prototype/issues/73), I added the resolvers and models for these, as well as corresponding unit tests.

Fixes # ([75](https://github.com/CDLUC3/dmsp_backend_prototype/issues/75))

## Type of change

## How Has This Been Tested?
These were tested by running the new unit tests, and also by testing the queries and mutations in Apollo Sandbox.

## Checklist:


src/models/MySqlModel.ts
```diff
@@ -199,9 +222,10 @@ export class MySqlModel {
     table: string,
     id: number,
     reference = 'undefined caller',
-  ): Promise<number> {
+  ): Promise<boolean> {
```
jupiter007, line 225: From what I observed, mySQL was not returning the data that was deleted. It was returning an object that looked like 
```
{
      fieldCount: 0,
      affectedRows: 1,
      insertId: 0,
      info: '',
      serverStatus: 2,
      warningStatus: 0,
      changedRows: 0
    }
```

src/models/MySqlModel.ts
```diff
@@ -169,7 +187,10 @@ export class MySqlModel {
   ): Promise<MySqlModel> {
     // Update the modifier info
     obj.modifiedById = apolloContext.token.id;
-    obj.modified = new Date().toISOString();
+    const now = new Date().toISOString();
+    const currentDate = now.slice(0, 19).replace('T', ' ');
```
jupiter007, line 191: mariaDB didn't seem to like the ISOString, so I had to add the `replace`

src/schemas/tag.ts
```diff
@@ -18,7 +19,7 @@ export const typeDefs = gql`
   "A Tag is a way to group similar types of categories together"
   type Tag {
     "The unique identifer for the Object"
-    id: Int!
+    id: Int
```
jupiter007, line 22: This "id" is auto-incremented in the database, so it's not required by the user.

src/schemas/section.ts
```diff
@@ -36,13 +38,13 @@ export const typeDefs = gql`
     "Indicates whether or not the section has changed since the template was last published"
     isDirty: Boolean!
     "The user who created the Object"
-    createdById: Int!
+    createdById: Int
     "The timestamp when the Object was created"
-    created: DateTimeISO!
+    created: String
```
jupiter007, line 43: Changed created and modified to String, to fix an issue mariadb had with ISO date type.

src/resolvers/versionedTemplate.ts
```diff
@@ -12,7 +12,7 @@ export const resolvers: Resolvers = {
     // Get all of the PublishedTemplates for the specified Template (a.k. the Template history)
     //    - called from the Template history page
     templateVersions: async (_, { templateId }, context: MyContext): Promise<VersionedTemplate[]> => {
-      if (isAdmin(context.token)){
+      if (isAdmin(context.token)) {
```
jupiter007, line 15: My "Prettier" extension added spaces here.

src/resolvers/tag.ts
```diff
@@ -0,0 +1,70 @@
...
+
+            if (await hasPermission(context, section.templateId)) {
+                return await Tag.getTagsBySectionId('tagsBySectionId resolver', context, sectionId);
+            }
+
+            throw ForbiddenError();
+        },
+    },
+    Mutation: {
+        addTag: async (_, { name, description }, context: MyContext): Promise<Tag> => {
```
jupiter007, line 29: I wasn't sure who has permissions to add and update tags, so I didn't add any checks for roles or permissions.

src/schemas/tag.ts
briri: Yes good catch.

src/resolvers/tag.ts
briri: I think this will only be SuperAdmin users, but it will never be allowed for Researchers so I think we can add a check for SuperAdmin for now.

briri (changes requested):
I've requested a few changes.

I'm also wondering if it would not be cleaner to just chain the SectionTag resolver to the Section resolver rather than doing the SQL `COALESCE` function since you have queries to fetch section both with and without the tags.

This will let GraphQL traverse to the Tags if it wants to. Something like:
```
export const resolvers: Resolvers = {
   // Queries
  // Mutations
  Section: {
    tags: async (parent: Section, _, context: MyContext): Promise<Tag[]> => {
      // A function that calls `SELECT tags.* FROM sectionTags JOIN tags ON sectionTags.tagId = tags.id WHERE sectionTags.sectionId = ?` 
      return await someQueryFunctionOnTheSectionTagsModel();
    }
}
```
It's also a good candidate to add chaining to allow someone querying for a Section to access the parent template (for example if we wanted to display the template title at the top of the page):
```
export const resolvers: Resolvers = {
   // Queries
  // Mutations
  Section: {
    // Allow the GraphQL client to fetch the template when querying for a Section
    template: async (parent: Section, _, context: MyContext): Promise<Template> => {
      return await Template.findById('Chained Section.template', context, parent.templateId);
    }
  }
```

You can then also in tun add one for the other direction by adding this to the `resolvers/template.ts`:
```
export const resolvers: Resolvers = {
   // Queries
  // Mutations
  Template: {
    // Allow the GraphQL client to fetch the template when querying for a Section
    sections: async (parent: Template, _, context: MyContext): Promise<Section[]> => {
      return await Section.findByTemplateId('Chained Template.sections', context, parent.templateId);
    }
  }
```

The places where you have `JSON_OBJECT` also seem like they would be better server with chaining resolvers.

src/models/MySqlModel.ts
```diff
@@ -148,10 +166,10 @@ export class MySqlModel {
                   (${props.map((entry) => entry.name).join(', ')}) \
                  VALUES (${Array(props.length).fill('?').join(', ')})`
 
-    const vals = props.map((entry) => this.prepareValue(entry.value));
+    const vals = props.map((entry) => this.prepareValue(entry.value, typeof (entry.value)));
 
     // Send the calcuated INSERT statement to the query function
-    const result = await this.query(apolloContext, sql.split(/[\s,\t,\n]+/).join(' '), vals, reference);
```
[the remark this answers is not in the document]
jupiter007: > Why remove this `split`? It handles the excess white space when a query is passed in that was written over multiple lines.

I think I remember it giving me errors. I'll add it back and check.

src/models/Section.ts
```diff
@@ -0,0 +1,198 @@
...
+                this.name,
+            );
+
+            // Then make sure it doesn't already exist
+            if (current) {
+                this.errors.push('Section with this name already exists');
+            } else {
+                /*Need to remove tags, because this does not exist in the sections table, but we need to
+                add tags into the response*/
+                delete this.tags;
```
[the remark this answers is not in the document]
jupiter007: Thanks for the suggestion. I will try that.

src/resolvers/section.ts
```diff
@@ -0,0 +1,186 @@
...
+                //Need to create a new instance of Section so that it recognizes the 'delete' function of that instance
+                const section = new Section({
+                    ...sectionData,
+                    id: sectionId
+                });
+
+                const deletedSection = await section.delete(context);
+
+                //Delete all sectionTags associated with this section
+                await SectionTag.deleteSectionTagsBySectionId('removeSection resolver', context, sectionId);
```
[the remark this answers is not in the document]
jupiter007: Ok thanks.

jupiter007:
> I've requested a few changes.
> 
> I'm also wondering if it would not be cleaner to just chain the SectionTag resolver to the Section resolver rather than doing the SQL `COALESCE` function since you have queries to fetch section both with and without the tags.
> 
> This will let GraphQL traverse to the Tags if it wants to. Something like:
> 
> ```
> export const resolvers: Resolvers = {
>    // Queries
>   // Mutations
>   Section: {
>     tags: async (parent: Section, _, context: MyContext): Promise<Tag[]> => {
>       // A function that calls `SELECT tags.* FROM sectionTags JOIN tags ON sectionTags.tagId = tags.id WHERE sectionTags.sectionId = ?` 
>       return await someQueryFunctionOnTheSectionTagsModel();
>     }
> }
> ```
> 
> It's also a good candidate to add chaining to allow someone querying for a Section to access the parent template (for example if we wanted to display the template title at the top of the page):
> 
> ```
> export const resolvers: Resolvers = {
>    // Queries
>   // Mutations
>   Section: {
>     // Allow the GraphQL client to fetch the template when querying for a Section
>     template: async (parent: Section, _, context: MyContext): Promise<Template> => {
>       return await Template.findById('Chained Section.template', context, parent.templateId);
>     }
>   }
> ```
> 
> You can then also in tun add one for the other direction by adding this to the `resolvers/template.ts`:
> 
> ```
> export const resolvers: Resolvers = {
>    // Queries
>   // Mutations
>   Template: {
>     // Allow the GraphQL client to fetch the template when querying for a Section
>     sections: async (parent: Template, _, context: MyContext): Promise<Section[]> => {
>       return await Section.findByTemplateId('Chained Template.sections', context, parent.templateId);
>     }
>   }
> ```
> 
> The places where you have `JSON_OBJECT` also seem like they would be better server with chaining resolvers.

Thanks @briri . That makes sense. I will work on adding the chaining.

jupiter007:
Hi Brian. I responded to your all your PR comments. Please take another look when you can. Thank you.

src/models/MySqlModel.ts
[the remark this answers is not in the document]
jupiter007: Brian, I ran tests after adding back the `sql.split(/[\s,\t,\n]+/).join(' ')` and it was giving me errors when I tried to insert or update the db table. So I kept it as is.

jupiter007:
> > I've requested a few changes.
> > I'm also wondering if it would not be cleaner to just chain the SectionTag resolver to the Section resolver rather than doing the SQL `COALESCE` function since you have queries to fetch section both with and without the tags.
> > This will let GraphQL traverse to the Tags if it wants to. Something like:
> > ```
> > export const resolvers: Resolvers = {
> >    // Queries
> >   // Mutations
> >   Section: {
> >     tags: async (parent: Section, _, context: MyContext): Promise<Tag[]> => {
> >       // A function that calls `SELECT tags.* FROM sectionTags JOIN tags ON sectionTags.tagId = tags.id WHERE sectionTags.sectionId = ?` 
> >       return await someQueryFunctionOnTheSectionTagsModel();
> >     }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It's also a good candidate to add chaining to allow someone querying for a Section to access the parent template (for example if we wanted to display the template title at the top of the page):
> > ```
> > export const resolvers: Resolvers = {
> >    // Queries
> >   // Mutations
> >   Section: {
> >     // Allow the GraphQL client to fetch the template when querying for a Section
> >     template: async (parent: Section, _, context: MyContext): Promise<Template> => {
> >       return await Template.findById('Chained Section.template', context, parent.templateId);
> >     }
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You can then also in tun add one for the other direction by adding this to the `resolvers/template.ts`:
> > ```
> > export const resolvers: Resolvers = {
> >    // Queries
> >   // Mutations
> >   Template: {
> >     // Allow the GraphQL client to fetch the template when querying for a Section
> >     sections: async (parent: Template, _, context: MyContext): Promise<Section[]> => {
> >       return await Section.findByTemplateId('Chained Template.sections', context, parent.templateId);
> >     }
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The places where you have `JSON_OBJECT` also seem like they would be better server with chaining resolvers.
> 
> Thanks @briri . That makes sense. I will work on adding the chaining.

Thanks for making it clear that I can use Resolver chaining. It made things way easier.

src/resolvers/section.ts
```diff
@@ -0,0 +1,193 @@
...
+
+                const updatedSection = await section.update(context);
+
+                // If there are errors than throw a Bad User Input error
+                if (updatedSection.errors) {
+                    const errorMessages = updatedSection.errors.join(', ');
+                    throw BadUserInput(errorMessages);
+                } else {
+                    //Get all the existing tags associated with this section
+                    const existingTags = await Tag.getTagsBySectionId('updateSection resolver', context, sectionId);
```
briri, line 133: I think all of the logic to figure out what tags are new could be moved to the new `sectionService`.

You are accounting for new tags that might have been added, but are we also looking for tags that might have been removed?

src/resolvers/versionedTemplate.ts
```diff
@@ -22,7 +22,7 @@ export const resolvers: Resolvers = {
     // Search for PublishedTemplates whose name or owning Org's name contains the search term
     //    - called by the Template Builder - prior template selection page
     publishedTemplates: async (_, { term }, context: MyContext): Promise<VersionedTemplate[]> => {
-      if (isAdmin(context.token)){
+      if (isAdmin(context.token)) {
         return await VersionedTemplate.search('publishedTemplates resolver', context, term);
       }
       // Unauthorized!
```
briri, line 28: Should add chaining here too go the versionedTemplate can get at all of the versionedSections

briri (changes requested):
Looks good @jupiter007 one more request to add chaining from the versionedTemplate to its versionedSections

and then moving the logic to handle tags into the sectionService and to also allow for tags to be removed from the list

src/resolvers/section.ts
jupiter007: @briri  - That makes sense to move the check for which new tags are not duplicates of existing tags to the sectionService. As far as removing tags, I was assuming when a tag is removed, it is completely removed from the database table. If that is the case, I don't think we need to check for tags that were removed.

src/resolvers/versionedTemplate.ts
jupiter007: That makes sense. Will update the schema for versionedTemplate to include versionedSection and add the resolver chaining.

jupiter007: Hi @briri I added chaining for versionedSection to versionedTemplate resolver, and also moved the logic for determining what new tags to add to sectionTags into sectionService.

src/resolvers/section.ts
briri: Ok. If you we can open a new ticket if we find any issues.

briri (approved):

src/resolvers/section.ts
jupiter007: Thanks Brian. I actually initially made the mistake of grabbing the tags from the tags table, rather than the sectionTags table. What I was trying to do was to get all the records from the secionTags table with the sectionId, and then prevent duplicate entries. I fixed that in the last commit :) 

I'm sure we'll encounter issues and areas of improvement as we start using the resolvers for the new pages. I will be sure to add new tickets accordingly.

Thanks again for your review.

Issue #48, opened 2024-07-19
Implement a logout endpoint and add ability to place tokens in a redis list as "invalidated tokens" 


jupiter007:
We did some research into the best approach for handling logouts securely. Ticket [#43 
](https://github.com/CDLUC3/dmsp_backend_prototype/issues/43). Please refer to the comments in that ticket for details on possible implementation.

The frontend will be deleting the httpOnly cookie that stores the JWT token when a user logs out. But, for added security, we want to be able to invalidate that token before that token expires.

The plan is that when a user logs out, the frontend will call the "/logout" endpoint in the backend and include the token info in the Authorization header. The backend will then grab the token, decode it, and grab the "JWT id" and calculate when that token expires. It will pass this info to be added to the invalid tokens list in redis.

We will need to create a function that maps to express-jwt's "isRevoked" property, so that it can be used to check whether the current token is invalid or not.

```
const authMiddleware = expressjwt({
  algorithms: ['HS256'],
  credentialsRequired: false,
  secret: generalConfig.jwtSecret as string,
  isRevoked: isRevokedCallback,
});
```

bofstein:
Will need to look into what Redis infrastructure we'll need for this.

PR #90, opened 2024-09-17
Support for refresh tokens and signing out

### LINKED ISSUE #81: Add "refresh token" to extend a user's active session
We need to add a "refresh token" to make sure that a user isn't prematurely logged out when actively using the site.

Here is some good documentation on implementing a "refresh token" when using "Jason Web Token" in NodeJS:
https://medium.com/@techsuneel99/jwt-authentication-in-nodejs-refresh-jwt-with-cookie-based-token-37348ff685bf

We can update our "verifyToken" function in tokenService.ts and authMiddleware.

briri:
## Description

- Added new endpoints for `/apollo-refresh` and `/apollo-signout` and `/apollo-csrf`
- CSRF tokens are added to the `X-CSRF-Token` header
- All non `GET`, `OPTIONS`, `HEAD` requests now require a CSRF token. 
  - The above `/apollo-csrf` endpoint can be used to fetch a CSRF token. 
  -  Each request to the backend (`/graphql`, `/apollo-refresh`, etc.) generate a new CSRF token.
- Updated Apollo endpoints so that the Access token is added as an HTTP-only cookie called `dmspt`
- The Refresh token is now added as an HTTP-only cookie called `dmspr`
- Access and Refresh tokens now contain their expiry info.
- Added a Redis cache to docker-compose 
- Added new ENV variables to the `.env.example` to support Redis, CSRF and JWT tokens
- Added tests for all controllers
- Added new Cache data source to interact with Redis
- Updated the tokenService to accommodate access tokens, refresh tokens, and the token black list
- Moved the authMIddleware function out of the `router.ts` file and into it's own middleware file
- Updated middleware and controllers to use the built-in express-jwt package's `auth` to access the decoded token
- Updated User model so that the bcrypt salt rounds are pulled from the ENV instead of hard-coded
- Updated README extensively
- Added diagrams of authentication logic to the README

Here is a summary of how auth works:
- Every request to the backend generates a CSRF token and adds it as an `X-CSRF-Token` header
- Every non `GET`, `OPTIONS` or `HEAD` request validates the CSRF token. A 403 `Invalid CSRF token` is returned if this fails
- User sign in/up generates an Access token set as the `dmspt` cookie and a Refresh token set as the `dmspr` token. The refresh token is added to the Redis cache
- Token refresh request generates both a new Access and Refresh token and updates the cookies and updates the refresh token in the Redis cache
- Sign out clears the cookies, deletes the Refresh token from the Redis cache and adds the Access token to a black list
- All tokens in the Redis cache are hashed to help prevent token manipulation

Fixes #81 and #48

## Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?

- Added Unit tests for all controllers 
- Updated the unit tests for the tokenService
- Added unit tests for the Cache datasource
- Added integration tests to ensure that the cache, headers and cookies are set as expected during sign in/up, refresh and sign out

Be sure to add/update the following ENV variables in your local `.env`. Use the ones in `.env.example` as reference
- BCRYPT_SALT_ROUNDS=10
- TOKEN_HASH_SECRET=your secret
- JWT_SECRET=ihef93hgf9-u3hgfi3hfte4g4tg4tg4
- JWT_TTL=600 _(ten minutes)_
- JWT_REFRESH_SECRET=7c94d41d-b01d-4a53-bef0-c818f254dbd8
- JWT_REFRESH_TTL=86400 _(1 day)_
- CSRF_LENGTH=32
- CSRF_TTL=3600 _(1 hour)_

## Checklist:


src/datasources/cache.ts
```diff
@@ -0,0 +1,50 @@
...
+  private constructor() {
+    // Setup the Redis Cluster
+    const cluster = new Redis.Cluster(cacheConfig.cluster);
+
+    // Having trouble figuring how how to type `Keyv` as `Keyv<string, Record<string, any>>`
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const keyV = new Keyv(new KeyvRedis(cluster)) as any;
+
+    keyV.on('connect', () => {
+      formatLogMessage(logger).error(null, `Redis connection established`);
```
jupiter007, line 21: L21 and L29: For these cases, would you want to use ".info()" instead of ".error()"?

briri: yes. good catch

src/controllers/__tests__/integrationTokens.spec.ts
```diff
@@ -0,0 +1,356 @@
...
+      .set('Content-Type', 'application/json')
+      .send(JSON.stringify(mockedUserData));
+
+    expect(resp.statusCode).toEqual(401);
+    expect(resp.headers['x-csrf-token']).toBeTruthy();
+    expect(resp.headers['set-cookie']).toBeFalsy();
+    expect(resp.body).toEqual({ success: false, message: 'Invalid credentials' });
+  });
+
+  it.only('POST /apollo-refresh should refresh the tokens', async () => {
```
jupiter007, line 258: L258: Only this unit test is running because of the "it.only". I removed it and tried running the unit test and the last test failed.

src/services/tokenService.ts
```diff
@@ -1,48 +1,227 @@
...
+export const generateAuthTokens = async (cache: Cache, user: User): Promise<{ accessToken: string; refreshToken: string }> => {
+  if (generalConfig.jwtSecret && generalConfig.jwtRefreshSecret && user && user.id && user.email) {
     try {
-      return jwt.sign(payload, generalConfig.jwtSecret as string, { expiresIn: generalConfig.jwtTtl });
+      // Generate a unique id for the JWT
+      const jti = new Date().getTime().toString();
+      // Generate an Access Token
+      const accessToken = generateAccessToken(jti, user);
+      // Generate a Refresh Token
+      const refreshToken = await generateRefreshToken(cache, jti, user.id);
```
jupiter007, line 114: L114 - Does the refresh token have to be refreshed every time a new access token is generated?

src/controllers/refreshTokenController.ts
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    const originalRefreshToken = req.cookies?.dmspr?.toString();
+
+    if (originalRefreshToken) {
+      const cache = Cache.getInstance();
+      const context = buildContext(logger, cache);
+      const { accessToken, refreshToken } = await refreshAuthTokens(cache, context, originalRefreshToken);
+
+      // If it successfully regenerated an access token
+      if (accessToken) {
```
jupiter007, line 18: L18 - Do we need to check for both the accessToken and the refreshToken before setting cookies for them?

src/index.ts
```diff
@@ -2,12 +2,15 @@ import { ApolloServer } from '@apollo/server';
 import express from 'express';
 import http from 'http';
 import cors from 'cors';
+import cookieParser from 'cookie-parser';
```
jupiter007, line 5: Looks like you're missing the "types" defs for cors and cookie-parser. Just need to run
 ```
npm i --save-dev @types/cors
 @types/cookie-parser
```

File: src/middleware/csrf.ts [hunk 6]
```diff
@@ -0,0 +1,20 @@
...
+import { generateCSRFToken, verifyCSRFToken } from '../services/tokenService';
+import { Cache } from '../datasources/cache';
+
+export async function csrfMiddleware(req: Request, res: Response, next: NextFunction) {
+  const cache = Cache.getInstance();
+
+  // Only worry about the CSRF token if the caller is performing a POST, PUT, PATCH, DELETE
+  if (req.method !== 'GET' && req.method !== 'HEAD' && req.method !== 'OPTIONS') {
+    const token = req.headers['x-csrf-token'] as string;
+    if (!token || !(await verifyCSRFToken(cache, token))) {
```
jupiter007, line 11: L11 - If the frontend initially calls the "/apollo-csrf" endpoint to get a CSRF token, will the user get a 'Invalid CSRF token' error instead of getting a new CSRF token?

src/controllers/__tests__/integrationTokens.spec.ts
briri: ok thanks. I will take a look

src/services/tokenService.ts
briri: I guess it depends. Do we want people to be logged in for 24 hours from the time that they sign in and then  have to sign in again or do we want to keep refreshing that token so that the countdown starts again?

src/controllers/refreshTokenController.ts
briri: yeah, that's probably a good idea

src/middleware/csrf.ts
briri: No. The "/apollo-csrf" endpoint is a GET, so the CSRF token verification doesn't get triggered

briri:
Thanks for noticing the `.only` on the integration test! I think I have addressed everything you pointed out in your review @jupiter007

jupiter007:
> Thanks for noticing the `.only` on the integration test! I think I have addressed everything you pointed out in your review @jupiter007

Thanks for making the updates Brian. It looks good. The only thing remaining is that when I run "npm test" I get 1 failing test with the error:

```
  ● Access and Refresh Tokens › POST /apollo-signout should remove cookies

    FetchError: request to https://auth.mydomain.edu/oauth2/token failed, reason: getaddrinfo ENOTFOUND auth.mydomain.edu

      at ClientRequest.<anonymous> (node_modules/node-fetch/lib/index.js:1501:11)

```

Do you get that same error? Maybe it's just in my environment?

briri:
weird I don't get any errors. The only place `oauth2/token` is referenced is in the DMPToolAPI class and its test. 

Did you run npm install?

jupiter007:
> weird I don't get any errors. The only place `oauth2/token` is referenced is in the DMPToolAPI class and its test.
> 
> Did you run npm install?

Yes I ran "npm install"

jupiter007 (approved):

jupiter007:
Thanks for the updates Brian. I approved the PR.

briri:
Redis cache has been built in AWS so going to merge and see how the pipeline works

PR #103, opened 2024-09-24
Chore/aws cache

briri:
Was able to setup SessionManager and log onto one of the ECS containers. From there I was able to connect to the Redis cache but had to use TLS.

This PR contains a modification to the config for the cache to use TLS when not in development mode. I also added a couple of debug lines (which I will remove once it's working).

### DIFF
--- a/src/config/cacheConfig.ts
+++ b/src/config/cacheConfig.ts
@@ -6,8 +6,10 @@ dotenv.config();
 verifyCriticalEnvVariable('CACHE_HOST');
 verifyCriticalEnvVariable('CACHE_PORT');
 
-export const cacheConfig = {
-  host: process.env.CACHE_HOST,
-  port: Number.parseInt(process.env.CACHE_PORT),
-  connectTimeout: Number.parseInt(process.env.CACHE_CONNECT_TIMEOUT) || 30000, // 30 seconds
-};
+const host = process.env.CACHE_HOST;
+const port = Number.parseInt(process.env.CACHE_PORT);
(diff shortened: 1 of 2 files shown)


src/config/cacheConfig.ts
```diff
@@ -6,8 +6,10 @@ dotenv.config();
...
-  port: Number.parseInt(process.env.CACHE_PORT),
-  connectTimeout: Number.parseInt(process.env.CACHE_CONNECT_TIMEOUT) || 30000, // 30 seconds
-};
+const host = process.env.CACHE_HOST;
+const port = Number.parseInt(process.env.CACHE_PORT);
+const connectTimeout = Number.parseInt(process.env.CACHE_CONNECT_TIMEOUT) || 30000; // 30 seconds
+
+export const cacheConfig = { host, port, connectTimeout };
+
+export const cacheTLS = `rediss://${host}:${port}`;
```
jupiter007, line 15: L15: Just checking that it's "rediss" with two "s"es.

Also, is it a good idea to mention that we use TLS for cache when not in development mode somewhere in the README?

jupiter007: Oh, and also, should you update the CHANGELOG?

briri: Yeah, was going to hold off on all that until I know these changes will work

jupiter007 (approved):

jupiter007:
Ok, approved the PR. Here's hoping it works :)

PR #113, opened 2024-09-30
Fixed bug with missing enum in VersionedQuestionCondition schema

jupiter007:
I noticed the bug after I merged my last PR into development.

The VersionedQuestionCondition schema is using an enum that is not defined in that schema


briri (approved):
👍🏻 I found this working on my current feature branch. Thanks for submitting a fix

PR #132, opened 2024-10-25
Added tests for Template versioning

### LINKED ISSUE #125: Update the unit tests for generating Versions
We should decide how we want to test the new functions for generating versioned templates, sections and questions.

The services unit tests need to be updated to add more coverage.

briri:
## Description

Fixes #125

- Added versioning tests for the TemplateService, SectionService and QuestionService
- Added integration test for the template versioning process. Intent is to ensure that the sections, questions and questionConditions are also versioned when calling `generateVersionedTemplate`. 

## Type of change

## How Has This Been Tested?

Ran the tests

Currently stuck with an async/timing issue in the `integrationVersioning.spec.ts`. The test uses local arrays and mockimplemepntations to get the code to store the results of the versioning process in the local arrays instead of a DB. That way we can inspect them to ensure that the foreign key relationships are correct and that changes made to the template, sections, questions and questionConditions are reflected the next time the template is versioned.

## Checklist:
- [ ] I updated the CHANGELOG.md and added documentation if necessary


jupiter007:
@briri, I ran a manual test of this branch on Apollo sandbox, and I saw some errors that I actually ran across and fixed in a previous branch. There are this types of conditions being used in questionService and sectionService, 
 ` if (created && created.errors?.length <= 0) {`
but that doesn't work because if `created.errors` is null or undefined, then `created.errors?.length<=0` will be undefined and the test will not pass.

I updated those instances and I got past some errors, but I am seeing some new ones. I will keep you posted.

jupiter007:
@briri I was able to get the versionedQuestionCondition table to be updated after fixing a couple of bugs:
1) Unfortunately, the word "condition" is a reserved word in mariaDB, which is a field name in versionedQuestionCondition table. So I temporarily added a workaround by replacing instances of "condition" with `condition` in order for the insert to work, but I think it would be better if we changed that field name.
2) Also, I find a typo in VersionedQuestionCondition model on line 13: "versionedQuestionConditions", needed to be updated to "versionedQuestionCondition";
3) I then tried to run the `integrationVersioning.spec.ts` unit test, but it was complaining that nothing was returned from " `await versionedQuestionCondition.create(context);`" in **questionService.ts**, so I think it's a mocking issue.

briri:
Oh great catch on the MariaDb condition reserved word! Thanks. I will update the data migration and the field in the model to something else.

Thanks for doing the research and the additional typos. Progress!

jupiter007:
Hi @briri , I noticed that I didn't check in some of my changes in my local copy of this branch:
- The questionService line "`if (created && created.errors?.length <= 0) {`" needs to be updated, because that condition will never be true if there is no "errors" object. 
- Also, on line 70 of sectionService, `"if (updated && updated.errors?.length <= 0) {`", the same thing

Sorry, for the late notice, I thought I checked those changes in.

briri:
thanks @jupiter007 I have updated them all. I also had to make some additional changes after doing some testing of the functionality and finding some bugs in the Apollo explorer

jupiter007:
Oh ok. Thanks @briri. I think your changes should be good to go, but I will take a quick look.

data-migrations/2024-09-25-1120-create-questions.sql
```diff
@@ -62,7 +62,7 @@ CREATE TABLE `versionedQuestions` (
   INDEX versionedQuestions_section_idx (`versionedSectionId`, `displayOrder`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8mb3;
 
-CREATE TABLE `versionedQuestionCondition` (
+CREATE TABLE `versionedQuestionConditions` (
```
jupiter007, line 65: @briri: I believe that "condition" needs to still be updated to conditionType on line 70.

jupiter007 (approved):

Issue #105, opened 2024-09-25
Backend table for alias emails


bofstein:

bofstein:
Create a table to be able to match Alias emails to a primary email for account management.  Will need ability to change primary email (through FE) as well

https://www.figma.com/proto/RRtoRLyr5Spf0KiyRfEclt/DMP-Explain?node-id=771-1054&t=LIKQ4JuadoMKIEAS-8&scaling=scale-down-width&content-scaling=fixed&page-id=736%3A1383&starting-point-node-id=736%3A1772&show-proto-sidebar=1&hotspot-hints=0&hide-ui=1

Need Schema and Resolvers as well.

Primary email is one that gets emails and can login.  The alias are only there for plan access if that's the email that is invited.

PR #136, opened 2024-10-31
Added resolvers for User mutations

### LINKED ISSUE #128: User updates needed for the profile page: Add language_id to the users table, and add a languages table, need updates to schema and resolvers to update user data
Currently, the new profile page needs a way to save the user's language, but there is no field for that under the users table

Wireframe: https://www.figma.com/proto/RRtoRLyr5Spf0KiyRfEclt/DMP-Tool-Wireframes?node-id=771-1054&t=LIKQ4JuadoMKIEAS-8&scaling=scale-down-width&content-scaling=fixed&page-id=736%3A1383&starting-point-node-id=736%3A1772&show-proto-sidebar=1&hotspot-hints=0&hide-ui=1

A graphql query to get the user data should also have the ability to grab the selection of languages available for our Language field drop-down.

Also, we need to add User mutation schemas and resolvers so that we can save the user's profile preferences.

Associated NextJS ticket to build profile page: https://github.com/CDLUC3/dmsp_frontend_prototype/issues/124

briri:
## Description

Fixes #128 and #105

- Added 2 new ENV variables: `DOMAIN` and `DEFAULT_AFFILIATION_URI`.
  - `DOMAIN` should equate to the URL of the UI. It will be used when anonymizing a user's email address and in links contained in email messages sent from the system.
  - `DEFAULT_AFFILIATION_URI` is used when a User account gets anonymized. It should equal the URI of your institution 
- Added data-migration to create the `userEmails` table, added new fields to `users` table and a query to set the `createdById` and `modifiedById` values on the seeded user entries.
- Updated the `register` function so that it sets the `createdById` and `modifiedById` to the current user after their account is created, adds the email to the `usersEmail` table.
- Updated the `emailService` to add placeholder functions for sending out the email confirmation messages and template collaboration notifications.
- Updated the TemplateCollaborator to call the new emailService function
- Updated the logic in the `User` model to work with the `UserEmail` model
- Added the `UserEmail` model. It's `create` function calls the emailService to send out a confirmation email
- Added `updateProfile`, `updateNotifications`, setUserOrcid` and `updatePassword` resolvers to support updates from profile pages
- Added `addUserEmail`, `removeUserEmail` and `setPrimaryUserEmail` resolvers to support adding secondary emails on profile pages
- Added `userService` that can `generateRandomPassword`, `anonymizeUser` and `mergeUsers`
- Added `removeUser` resolver that will anonymize the user's account. We will need to keep this up to date as we add new tables
- Added some initial Admin resolvers for managing users: `deactivateUser`, `activateUser` which can be used to lock an account and `mergeUsers` which can be used to consolidate duplicate accounts.
- Added `recordLogIn` to the User model which will now update a `last_sign_in_at` timestamp and `last_sign_in_by` (`PASSWORD` or `SSO`) to help us troubleshoot user authentication issues.

Also fixed:
- An issue on the `Affiliation` model that was causing JSON data in the DB to be returned as a Sring instead of properly parsed into Arrays
- Updated resolver chain references to `Affiliation` to use the `Affiliation.uri` instead of `Affiliation.id`
- Some issues where `password` was being validated but we're working with the hashed version so it's not appropriate to run the validation check.
- Added `languageId` to `VersionedTemplate`. Added it to `Template` on a prior PR but forgot this one
- Rename the GraphQL error `BadUserInput` to `BadUserInputError` to follow the name pattern of others
- Fixed issue with the `isAdmin` auth check and the affiliation info
- Added missing `helpers` tests

## Type of change

## How Has This Been Tested?

Added unit tests and also tested within the Apollo server explorer.

Note that you need to add the 2 new ENV variables to your dot env file!

You should also run the scripts to nuke and rebuild your local DB

## Checklist:


src/resolvers/user.ts
```diff
@@ -1,36 +1,322 @@
...
+      notify_on_comment_added,
+      notify_on_template_shared,
+      notify_on_feedback_complete,
+      notify_on_plan_shared,
+      notify_on_plan_visibility_change,
+    } }, context: MyContext): Promise<User> => {
+      if (isAuthorized(context?.token)) {
+        const user = await User.findById('updateUserNotifiactions resolver', context, context.token.id);
+        // Only continue if the user is active and not locked
+        if (!user || !user.active || user.locked) {
```
jupiter007, line 88: L88: Not a big deal, but I noticed this is used multiple times, and was wondering if it'd make sense to put it in a separate function?

src/resolvers/user.ts
```diff
@@ -1,36 +1,322 @@
...
+          if (!updated) {
+            throw InternalServerError('Unable to activate the user at this time');
+          }
+          return result as User;
+        }
+      }
+      // Unauthorized!
+      throw context?.token ? ForbiddenError() : AuthenticationError();
+    },
+    // Merge the 2 user accounts (Admin only)
```
jupiter007, line 289: L289: Should the comment also include "SuperAdmin", or is it understood that "Admin only" captures both Admin and SuperAdmin?

jupiter007:
Brian, I'll plan on taking a closer look at the code tomorrow, and run some tests, but it looks good so far.

src/resolvers/user.ts
briri: Yes. I am going to create a ticket to add some helper methods for some duplicative code. I was thinking about it recently. Like our `if (updated && (updated.errors && Array.isArray(updated.errors) ...`

jupiter007: Thanks Brian.

src/resolvers/user.ts
```diff
@@ -1,36 +1,322 @@
...
+      if (isAdmin(context.token)) {
+        const userToMerge = await User.findById('mergeUsers resolver', context, userIdToBeMerged);
+        const userToKeep = await User.findById('mergeUsers resolver', context, userIdToKeep);
+        // Only continue if the current user's affiliation matches the user OR they are SuperAdmin
+        const affil = context.token.affiliationId;
+        if (
+          (affil === userToMerge.affiliationId && affil === userToKeep.affiliationId) ||
+          isSuperAdmin(context.token)
+        ) {
+          const merged = await mergeUsers(context, userToMerge, userToKeep);
```
jupiter007, line 300: L300: When you pass `userToKeep` into `mergeUsers`, I think you have to create an instance of the "User" class for that, because when I try and merge two users, I get an error that "**userToKeep.update is not a function**", which is reminiscent of the other cases where I had to create a new instance of a class in order to access its methods.

src/models/UserEmail.ts
```diff
@@ -0,0 +1,194 @@
...
+      }
+    } else {
+      this.errors.push('Email has not been created yet');
+    }
+
+    return this;
+  }
+
+  //Delete this UserEmail
+  async delete(context: MyContext): Promise<UserEmail> {
```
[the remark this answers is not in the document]
briri: No I don't think so, that would prevent a user from deleting it if for example they added it and had a typo and could never confirm it.

src/resolvers/user.ts
```diff
@@ -1,36 +1,322 @@
...
+        const user = await User.findById(ref, context, context.token.id);
+        // Only continue if the user is active and not locked
+        if (!user || !user.active || user.locked) {
+          throw ForbiddenError();
+        }
+
+        const userEmails = await UserEmail.findByUserId(ref, context, context.token.id);
+
+        const existing = userEmails.find((entry) => { return entry.email === email });
+        const oldPrimary = userEmails.find((entry) => { return entry.isPrimary === true });
```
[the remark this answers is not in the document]
briri: ah ok interesting. It didn't give me any issues. I will update the table definition

src/models/User.ts
```diff
@@ -224,4 +294,58 @@ export class User extends MySqlModel {
       return this;
     }
   }
+
+  // Save the changes made to the User
+  async update(context: MyContext): Promise<User> {
```
jupiter007, line 299: Not a big issue, but I noticed that additional tests weren't added for the User model, even with all these changes. So the coverage for User.ts is pretty low. But we can address that at a later time.

src/resolvers/user.ts
```diff
@@ -1,36 +1,322 @@
...
+        const user = new User(result);
+        // Only continue if the current user's affiliation matches the user OR they are SuperAdmin
+        if (context.token.affiliationId === user.affiliationId || isSuperAdmin(context.token)) {
+          user.active = true;
+          const updated = await User.update(context, user.tableName, user, ref, ['password']);
+
+          if (!updated) {
+            throw InternalServerError('Unable to activate the user at this time');
+          }
+          return result as User;
```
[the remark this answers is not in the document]
briri: I was wondering why it was forcing me to add the `as User` part. I didn't realize it was returning the object with the `affectedRows` property.

src/resolvers/user.ts
briri: I am wondering why this happens and think maybe the issue is that the `MySQLModel.query` returns its result(s) which contains all of the properties but not the functions that Model class has, so I think I was getting confused because `user.email` works but `user.update` doesn't.

I will add a note to that new cleanup ticket to update this. It will be good to update all of those functions so that we return an instance of the model instead of just the query result

jupiter007: Yea, it would be good to create a ticket to investigate that.

src/models/User.ts
briri: Oh haha. I left an `it.only(` in place. I just removed it and made changes for the other issues you mentioned.

jupiter007: Thanks Brian.

src/models/User.ts
```diff
@@ -341,7 +340,12 @@ export class User extends MySqlModel {
         this.password = await this.hashPassword(newPassword);
 
         const updated = await User.update(context, this.tableName, this, 'User.updatePassword');
-        return updated as User;
+
+console.log(updated)
```
jupiter007, line 344: L344: Did you mean to leave the console.log here?

src/resolvers/user.ts
jupiter007: Maybe make a note here or in #137 that `mergeUsers` needs to be fixed due to this error, just so we don't forget that this function isn't working and needs to be updated.

briri: Oh, I thought I had fixed the issue. I'll add a note to the ticket

jupiter007: Ok thanks Brian.

jupiter007 (approved):
Great work Brian. Thanks for the updates.

Issue #141, opened 2024-11-05
Fix some bugs in user resolver to accommodate updates in User Profile page


jupiter007:
As I was starting to build the User Profile page ([#124](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/124)), and connect it to the backend data, I noticed some bugs.

PR #142, opened 2024-11-05
Made some updates to user resolver while using the user queries and mutations for the user profile page

jupiter007:
## Description

While working on the User Profile page, noticed a few bugs while using the user queries and mutations. This ticket includes those updates.

- Update user resolver
-- Updated the "me" query to `findById` rather than `findByEmail`, since data was disappearing on the user profile page when a user updated their email preference.
-- Switched to returning `created` in `addUserEmail` function, instead of `userEmail`, so that errors would be returned to the client
-- Updated `removeUserEmail` to return the data with any possible errors
-- Updated `setPrimaryUserEmail` to only update `oldPrimary` if there were no errors when making the `set primary` update. Also, if there are errors, we should return the original data, not the updated data, with the errors. What was happening was that the user was setting an email to be primary, and the backend response returned data showing that that email's isPrimary was true, even when the change wasn't made. So the page would falsely reflect that update.
- Updated UserEmail  model
-- Updated isValid to use the shared "validatedEmail" function

Fixes # (#141 )

## Type of change

## How Has This Been Tested?
I ran unit tests, and manually tested via Apollo Sandbox/Explorer. I also tested with the User Profile page which is still in development.

## Checklist:

### DIFF
--- a/src/models/UserEmail.ts
+++ b/src/models/UserEmail.ts
@@ -1,5 +1,6 @@
 import { MyContext } from "../context";
 import { sendEmailConfirmationNotification } from "../services/emailService";
+import { validateEmail } from '../utils/helpers';
 import { TemplateCollaborator } from "./Collaborator";
 import { MySqlModel } from "./MySqlModel";
 
@@ -25,13 +26,13 @@ export class UserEmail extends MySqlModel {
   // Validation to be used prior to saving the record
   async isValid(): Promise<boolean> {
     await super.isValid();
-
     if (this.userId === null) {
       this.errors.push('User can\'t be blank');
     }
-    if (!this.email) {
-      this.errors.push('Email can\'t be blank');
+    if (!validateEmail(this.email)) {
+      this.errors.push('Enter valid email');
     }
+
     return this.errors.length <= 0;
   }
 
@@ -94,7 +95,7 @@ export class UserEmail extends MySqlModel {
         this.errors.push('Email has already been confirmed by another account');
       }
 
-      if (this.errors.length <= 0){
+      if (this.errors.length <= 0) {
         // Save the record and then fetch it
         const newId = await UserEmail.insert(context, this.tableName, this, ref);
         const created = await UserEmail.findById(ref, context, newId);
--- a/src/resolvers/user.ts
+++ b/src/resolvers/user.ts
@@ -13,7 +13,7 @@ export const resolvers: Resolvers = {
     // returns the current User
     me: async (_, __, context: MyContext): Promise<User> => {
       if (isAuthorized(context?.token)) {
-        return await User.findByEmail('me resolver', context, context.token.email);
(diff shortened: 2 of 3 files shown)


src/resolvers/user.ts
```diff
@@ -13,7 +13,7 @@ export const resolvers: Resolvers = {
     // returns the current User
     me: async (_, __, context: MyContext): Promise<User> => {
       if (isAuthorized(context?.token)) {
-        return await User.findByEmail('me resolver', context, context.token.email);
+        return await User.findById('me resolver', context, context.token.id);
```
jupiter007, line 16: Updated the "me" query to findById rather than findByEmail, since data was disappearing on the user profile page when a user updated their email preference.

src/resolvers/user.ts
```diff
@@ -176,9 +176,9 @@ export const resolvers: Resolvers = {
           throw NotFoundError();
         }
 
-        const original = structuredClone(userEmail);
-        if (await new UserEmail(userEmail).delete(context)) {
-          return original;
+        const deleted = await new UserEmail(userEmail).delete(context);
+        if (deleted) {
+          return deleted; //Return the deleted email with any errors from calling delete function
```
jupiter007, line 181: Updated removeUserEmail to return the data with any possible errors

src/resolvers/user.ts
```diff
@@ -196,27 +196,35 @@ export const resolvers: Resolvers = {
...
 
-        if (oldPrimary) {
-          oldPrimary.isPrimary = false;
-          await new UserEmail(oldPrimary).update(context);
-        }
-
         existing.isPrimary = true;
         const updated = await new UserEmail(existing).update(context);
         if (updated && (!updated.errors || (Array.isArray(updated.errors) && updated.errors.length === 0))) {
+          // Update old primary record to isPrimary = false, if the new one was updated successfullly
```
jupiter007, line 210: Updated setPrimaryUserEmail to only update oldPrimary if there were no errors when making the set primary update. Also, if there are errors, we should return the original data, not the updated data, with the errors. What was happening was that the user was setting an email to be primary, and the backend response returned data showing that that email's isPrimary was true, even when the change wasn't made. So the page would falsely reflect that update.

src/models/UserEmail.ts
```diff
@@ -25,13 +26,13 @@ export class UserEmail extends MySqlModel {
   // Validation to be used prior to saving the record
   async isValid(): Promise<boolean> {
     await super.isValid();
-
     if (this.userId === null) {
       this.errors.push('User can\'t be blank');
     }
-    if (!this.email) {
-      this.errors.push('Email can\'t be blank');
+    if (!validateEmail(this.email)) {
```
jupiter007, line 32: Updated isValid to use the shared "validatedEmail" function

PR #144, opened 2024-11-15
Allow for affiliation creation when one is not selected from the typeahead

### LINKED ISSUE #138: Make updates to updateUserProfile schema and resolver to accommodate the "Other" affiliation scenario
Currently, on the User Profile form's Institution dropdown, there is an option to select "Other" and add a custom institution name. However, there is no way to create a new Affiliation because the current schema for adding an affiliation has other required fields, like "funder" and "uri", etc.

Currently, there is some logic in place to use default values:
```
if (!this.uri) {
  this.uri = `${DEFAULT_DMPTOOL_AFFILIATION_URL}${randomHex(6)}`;
}
this.funder = options.funder || false;
this.active = options.active || false;
```
but the schema needs to be updated to make those required fields optional.

It might make sense to allow the client to just submit an "OtherAffiliation" field, and have the backend handle the creation of the new Affiliation and assignment of the new id to the updated user profile data.

briri:
## Description

Fixes #138 

- Added `otherAffiliationName: String` to the `updateUserProfile` resolver
- Also allowing `otherAffiliationName: String` to be passed to the `apollo-signup` endpoint
- Created `affiliationService` that has a method that will find the name specified. If it exists, it will use that `Affiliation` otherwise it will create a new one and place the new `affiliationId` on the `User`
- Added `findByName` to the Affiliation model and fixed some minor issues with the serialization of Arrays 
- Noticed that the `MySQLModel.prepareValue` function was getting `object` instead of `Object` for the type, so added handlers for `object` and `array`

## Type of change

## How Has This Been Tested?

Added unit tests for the new affiliationService and updated existing tests.

Tested the `updateUserProfile` mutation manually via the Apollo explorer http://localhost:4000/graphql

@jupiter007 the frontend will need to pass the other affiliation text as `otherAffiliationName` when updating from the Profile page as well as the signup form. I'm not sure if that is something you set in the typeahead component or the individual pages.

## Checklist:

### DIFF
--- a/src/models/MySqlModel.ts
+++ b/src/models/MySqlModel.ts
@@ -64,13 +64,17 @@ export class MySqlModel {
     if (val === null || val === undefined) {
       return null;
     }
-
     switch (type) {
       case 'number':
         return Number(val);
       case 'json':
         return JSON.stringify(val);
+      case 'object':
+        return JSON.stringify(val);
       case Object:
+        return JSON.stringify(val);
+      case 'array':
+        return JSON.stringify(val);
       case Array:
         return JSON.stringify(val);
       case 'boolean':
@@ -89,6 +93,7 @@ export class MySqlModel {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   static propertyInfo(obj: Record<string, any>, skipKeys: string[] = []): { name: string, value: string }[] {
     const excludedKeys = ['id', 'errors', 'tableName'];
+
     return Object.keys(obj)
       .filter((key) => ![...excludedKeys, ...skipKeys]
         .includes(key)).map((key) => ({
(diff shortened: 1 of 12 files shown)


jupiter007 (approved):
LGTM

src/models/MySqlModel.ts
```diff
@@ -64,13 +64,17 @@ export class MySqlModel {
     if (val === null || val === undefined) {
       return null;
     }
-
     switch (type) {
       case 'number':
         return Number(val);
       case 'json':
         return JSON.stringify(val);
+      case 'object':
```
jupiter007, line 72: L72: @briri this is breaking /login because a date is of type 'object' and it is being converted into a string, instead of being handled by the `default` case.

jupiter007: Thanks @briri. LGTM.

PR #145, opened 2024-11-18
Bug/affiliations table definition

briri:
## Description

Found a small bug while trying to run the latest data-migrations in the AWS environment. MySQL does not allow default values for JSON, TEXT, BLOB, etc. data types.

## Type of change

## How Has This Been Tested?

Successfully ran the data-migrations in the AWS dev environment


jupiter007 (approved):
LGTM

PR #148, opened 2024-12-05
Fixed issue with Cache

briri:
## Description

Fixed the issue we were having with the AWS Elasticache. It turns out that dealing with a ValKey/Redis cache in cluster mode requires keys to be defined a little differently.

The bug/issue arose when I switched over from running the serves Redis instance to a ValKey cache cluster

- In a Redis cluster, keys are distributed across multiple node groups (shards) based on a hash slot.
- ioredis (the package we use) automatically handles the distribution of keys across shards, but we need to be mindful of how keys are hashed to ensure even distribution and avoid cross-slot operations.
- Redis clusters do not support commands that operate on keys located in different hash slots.
- To group related keys (e.g., for CSRF tokens), we need to use hash tags (`{}`) to force certain keys to the same hash slot. For example `{csrf}:7836824iugu2h4g425t23t` instead of `csrf:7836824iugu2h4g425t23t`

## Type of change

## How Has This Been Tested?

I deployed to AWS and was able to login again. I also verified that all of the unit and integration tests are still passing

## Checklist:

### DIFF
--- a/src/services/tokenService.ts
+++ b/src/services/tokenService.ts
@@ -51,7 +51,7 @@ export const generateCSRFToken = async (cache: Cache): Promise<string> => {
     const hashedToken = await hashToken(csrfToken);
 
     // Add the refresh token to the Cache
-    await cache.adapter.set(`csrf:${csrfToken}`, hashedToken, { ttl: generalConfig.csrfTTL });
+    await cache.adapter.set(`{csrf}:${csrfToken}`, hashedToken, { ttl: generalConfig.csrfTTL });
     return csrfToken;
   } catch(err) {
     formatLogMessage(logger).error(err, 'generateCSRFToken error!');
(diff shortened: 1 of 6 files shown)


jupiter007 (approved):
Glad you discovered the issue. Great work.

src/services/tokenService.ts
```diff
@@ -51,7 +51,7 @@ export const generateCSRFToken = async (cache: Cache): Promise<string> => {
     const hashedToken = await hashToken(csrfToken);
```
jupiter007, line 51: L51: I don't think you need to await `hashToken()` do you? It's not returning a promise from what I can tell.

briri: good catch, I have updated to remove that unnecessary await.

briri:
looks like this is working. Have re-deployed a bunch of times and its still functioning.

PR #168, opened 2025-01-09
Additional fixes for the signup logic

briri:
@andrewebdev I was able to successfully create an account with an "other" affiliation with these additional changes.

- I incorporated the changes @jupiter007 identified in the comments on PR #163.
- I pulled in the latest changes from the development branch (This is the addition of the custom `insert` logic on the `UserEmail` model)
- @jupiter007 I ended up changing 
  - the `User.findById` so that it does the `new User(user)` so the caller always has access to the class functions. I will create a separate ticket to go back through the code and clean up these finders and functions like register so that they all have the same behavior and return the class with all of its functions instead of just the properties.
  - I added a check to MySQLModel's `update` function to handle this scenario where the token has no user id.

@andrewebdev if you merge this into your branch we should be able to get it merged into development today


jupiter007:
Thanks @briri. @andrewebdev let me know once you merge and I can run another quick test.

Issue #173, opened 2025-01-14
We need a query to get ALL public templates, even ones outside of user's affiliation


jupiter007:
The `Select existing template` page in the NextJS app (located at /template/create), will need to be able to grab ALL public templates, regardless of whether they are within user's affiliation. 

Currently, there is no query in the backend to do this. We need to add another query for public templates.

PR #174, opened 2025-01-14
Updates to accommodate Create Template start pages

jupiter007:
## Description

For the `Select existing template` page of the Create Template start pages, we need a way to 

1. Pull all versionedTemplates (i.e. published templates) that have a visibility=PUBLIC
2. Pull all versionedTemplates that are under the user's affiliation

- Added `findPublicTemplates` and `findByUser` queries to VersionedTemplate model
- Updated `addTemplate` method in resolver to clone template-related sections and questions when a user wants to create a new template off of an existing one
- Updated `versionedTemplate` schema with the addition of `userAffiliationTemplates` and `publicVersionedTemplates` and added those new queries to the versionedTemplate resolver.
- added unit tests for VersionedTemplate model

Fixes # ([173](https://github.com/CDLUC3/dmsp_backend_prototype/issues/173))

## Type of change

## How Has This Been Tested?
Added unit test and also tested manually using Apollo Explorer

## Checklist:


jupiter007:
I put this back into `DRAFT` because this branch may need more updates to work as expected with the Create Template start pages.

src/models/VersionedTemplate.ts
```diff
@@ -121,4 +122,25 @@ export class VersionedTemplate extends MySqlModel {
     const results = await VersionedTemplate.query(context, sql, [versionedTemplateId.toString()], reference);
     return Array.isArray(results) && results.length > 0 ? results[0] : null;
   }
+
+  //Find all versionedTemplates that have a visibility = PUBLIC
```
briri, line 126: I don't think either of these are necessary. I think I missed the context it.

The versionedTemplates resolver has a `search` function which in turn calls the `publishedTemplates` function on the VersionedTemplate model. It returns only "published" templates. If we need to add a flag in to restrict it to only "public" templates that would be simple enough. 

The `templates` resolver has a `templates` function which in turn calls the `findByUser` function on the Template model that will return the latest data for all of the templates owned by the current user's affiliation or that they are a collaborator on.

briri (commented):
I'm still trying to catch up this morning 🙃 but I don't think we need these new queries

src/models/VersionedTemplate.ts
jupiter007: @briri , my understanding is that for the "Start with a copy of an existing template" page, when a user first arrives on the page, we have to pull all published templates for the user's org without a `search` term, and pull all `public` published templates without a `search` term. Then when a user enters a search term on the page, we just filter it on the frontend. The user can see all templates without specifying a search term.

briri: ah ok, I guess if we get feedback asking to be able to copy an unpublished or previously published template we can modify the query later.

Anyway if we're restricting it to only published public templates, the `publishedTemplates` mutation in the `VersionedTemplate` resolver should give you everything you need. You could then in the frontend filter to extract the ones whose `affiliationId` matches the one for the current user to get the list of "my templates" and then filter on `visibility` public and `affiliationId` NOT equal to the current user's  to get all of the rest.

jupiter007: @briri I updated the schema for `publishedTemplates` so that it doesn't require a "search" term, and that then can be used in place of the `publicVersionedTemplates` query that I added. So I will remove `publicVersionedTemplates` from the code. However, it looks like we'd want to keep the `userAffiliationTemplates` query that I added, because, unlike `templateVersions` query, it's not searching on `templateId`. We might want to keep both queries. What do you think?

jupiter007: @briri I updated the branch to remove `publicVersionedTemplates` query and update the existing `publishedTemplates` query so that it will not required the "search" term, and also, so that it will not place `undefined` in as a search term when it is not present. 

As discussed, I need to keep the `userAffiliationTemplates` query still, so I left it in place.

Please take another look when you get the chance.

src/models/VersionedTemplate.ts
```diff
@@ -121,4 +124,18 @@ export class VersionedTemplate extends MySqlModel {
     const results = await VersionedTemplate.query(context, sql, [versionedTemplateId.toString()], reference);
     return Array.isArray(results) && results.length > 0 ? results[0] : null;
   }
+
+  // Find all of the templates associated with the context's User's affiliation
+  static async findByUser(reference: string, context: MyContext): Promise<VersionedTemplate[]> {
```
briri, line 129: Can we switch the name on this to be more literal and pass in the affiliationId explicitly? This will make it more reusable when we want to create some administrative pages.

Something like:
``` 
static async findByAffiliationId(reference: string, context: MyContext, affiliationId: string): Promise<VersionedTemplate[]> {
 ...
}
```
and then in the resolver call like:
``` 
const affiliationTemplates = await Versionedtemplate. findByAffiliationId(reference, context, context.token?.affiliationId);
```

src/models/VersionedTemplate.ts
```diff
@@ -121,4 +124,18 @@ export class VersionedTemplate extends MySqlModel {
...
+    const sql = 'SELECT * FROM versionedTemplates WHERE ownerId = ? ORDER BY modified DESC';
+    const templates = await VersionedTemplate.query(context, sql, [context.token?.affiliationId], reference);
+
+    // Also look for any templates that the current user has been invited to collaborate on
+    const sharedTemplates = await TemplateCollaborator.findByEmail(
```
briri, line 134: I don't think you need to have this `sharedTemplates` query at all. TemplateCollaborator is associated with Template not VersionedTemplates and will return a different object than the query above.

I think it would be pretty edge-case for a user to want to copy a non-public published template that they were invited to collaborate on. If we need to come back and do that later on we can, but we'll need to write a fairly complex join query to get at the current versioned template

src/resolvers/template.ts
```diff
@@ -44,26 +48,53 @@ export const resolvers: Resolvers = {
...
+        let template: Template;
         if (copyFromTemplateId) {
           // Fetch the VersionedTemplate we are cloning
           const original = await VersionedTemplate.findVersionedTemplateById(
             'addTemplate resolver',
             context,
             copyFromTemplateId
           );
-          template = await cloneTemplate(context.token?.id, context.token.affiliationId, original);
+          template = cloneTemplate(context.token?.id, context.token.affiliationId, original);
```
briri, line 59: good catch on the unnecessary `await` here

src/resolvers/versionedTemplate.ts
```diff
@@ -29,6 +29,14 @@ export const resolvers: Resolvers = {
       // Unauthorized!
       throw context?.token ? ForbiddenError() : AuthenticationError();
     },
+    // Get the VersionedTemplates that belong to the current user's affiliation (user must be an Admin)
+    userAffiliationTemplates: async (_, __, context: MyContext): Promise<VersionedTemplate[]> => {
```
briri, line 33: can we change this one to be `myVersionedTemplates`? 

I have just created a `myProjects` resolver and will be changing the `templates` resolver to `myTemplates` so that it's clearer what is being fetched.

briri (commented):
Looks good. Just a few minor changes

src/models/VersionedTemplate.ts
jupiter007: Makes sense. I like that name better as well. I made the update.

src/models/VersionedTemplate.ts
jupiter007: Thanks for confirming Brian. I wasn't sure whether we needed it or not. I removed it.

src/resolvers/versionedTemplate.ts
jupiter007: That makes sense. I like the consistency of `myVersionedTemplates`. Sometimes naming things can be the toughest :)

jupiter007:
Thanks so much for your review @briri. I made the updates you requested.

briri (approved):
Thanks for making those changes @jupiter007!

Issue #193, opened 2025-02-06
Make some more updates to the backend code to accommodate Edit/Add Question form


jupiter007:
There were still some updates that needed to be made to backend code so that Edit/Add Question functionality would work for Options.

These changes were for the https://github.com/CDLUC3/dmsp_frontend_prototype/issues/188 Frontend work.

PR #194, opened 2025-02-06
Updates required for the https://github.com/CDLUC3/dmsp_frontend_prototype/issues/188 frontend Edit Question ticket

jupiter007:
## Description

- Removed `uniq` constraints on `text` and `orderNumber` for questionOptions table, because it was restricting the ability to update any text on an existing record.
- Added `getQuestionOptionsToRemove` to remove questionOptions that are no longer in the updated set
- Updated `questionOptions` resolver to use `id` instead of `questionOptionId` because otherwise I had to extend existing class

Fixes # ([193](https://github.com/CDLUC3/dmsp_backend_prototype/issues/193))

## Type of change

## How Has This Been Tested?
Manual testing and also checked that it didn't break any existing unit tests

## Checklist:
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/data-migrations/2025-01-24-0318-create-question-options-table.sql
+++ b/data-migrations/2025-01-24-0318-create-question-options-table.sql
@@ -8,7 +8,5 @@ CREATE TABLE `questionOptions` (
   `createdById` int,
   `modified` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `modifiedById` int,
-  CONSTRAINT unique_question_option_text UNIQUE (`questionId`, `text`),
-  CONSTRAINT unique_question_option_orderNumber UNIQUE (`questionId`, `orderNumber`),
   FOREIGN KEY (questionId) REFERENCES questions(id) ON DELETE CASCADE
 ) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8mb3;
--- a/src/resolvers/question.ts
+++ b/src/resolvers/question.ts
@@ -140,15 +140,33 @@ export const resolvers: Resolvers = {
             existingQuestionOptions.map(option => [option.id, option])
           );
 
+
+          // Get list of options that need to be removed
+          const optionsToRemove = await getQuestionOptionsToRemove(questionOptions as QuestionOption[], context, questionId);
+
+          // Remove question options that are no longer in the updated questionOptions array
+          if (optionsToRemove.length > 0) {
+            await Promise.all(
+              optionsToRemove.map(async (option) => {
+                const questionOption = new QuestionOption({
+                  questionId: option.questionId,
+                  id: option.id
+                });
+
+                await questionOption.delete(context);
+              })
+            );
+          }
+
           // Separate incoming options into "to update" and "to create"
           const optionsToUpdate = [];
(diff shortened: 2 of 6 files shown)


data-migrations/2025-01-24-0318-create-question-options-table.sql
```diff
@@ -8,7 +8,5 @@ CREATE TABLE `questionOptions` (
   `createdById` int,
   `modified` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   `modifiedById` int,
-  CONSTRAINT unique_question_option_text UNIQUE (`questionId`, `text`),
```
jupiter007, line 11: I had to remove these constraints because they were preventing me from updating existing question options on the Edit Question page

src/resolvers/question.ts
```diff
@@ -3,7 +3,7 @@ import { MyContext } from "../context";
 import { QuestionOption } from "../models/QuestionOption";
 import { Question } from "../models/Question";
 import { Section } from "../models/Section";
-import { hasPermissionOnQuestion } from "../services/questionService";
+import { getQuestionOptionsToRemove, hasPermissionOnQuestion } from "../services/questionService";
```
jupiter007, line 6: Needed to add getQuestionOptionsToRemove function for when a user removes options from an existing questions.

src/resolvers/question.ts
```diff
@@ -140,15 +140,33 @@ export const resolvers: Resolvers = {
...
+            );
+          }
+
           // Separate incoming options into "to update" and "to create"
           const optionsToUpdate = [];
           const optionsToCreate = [];
 
           questionOptions.forEach(option => {
-            if (existingOptionsMap.has(option.questionOptionId)) {
+            if (existingOptionsMap.has(option.id)) {
```
jupiter007, line 166: Switched to using `id` instead of `questionOptionId` so that I didn't need to extend the existing QuestionOption class

src/resolvers/question.ts
briri: I have a list of new tickets I am going to create and consolidating how associations like this are handled is one of them. 

Let's merge this fix for now and we can come back afterward to make sure we're handling them all in the same way.

src/resolvers/question.ts
briri: yeah that makes sense. You're working within the context of the QuestionOption itself here.

src/resolvers/questionOption.ts
```diff
@@ -74,22 +74,5 @@ export const resolvers: Resolvers = {
         return await QuestionOption.findByQuestionOptionId('updateQuestion resolver', context, updatedQuestionOption.id);
       }
     },
-    removeQuestionOption: async (_, { questionOptionId }, context: MyContext): Promise<QuestionOption> => {
```
briri, line 77: I was also thinking about situations like this. There are a few cases where we exposed individual `removeThing` resolvers, but the front end sends the list of its associations when the parent object gets created/updated. 

As part of the ticket I mentioned above, we should probably figure out which approaches we are using for each scenario and delete any unused ones

briri (approved):
LGTM. I was to do some consolidation of how we handle these association relationships but we can do that in a separate ticket

Issue #210, opened 2025-02-22
[Bugs] - Make bug fixes to allow Template Builder pages to work


jupiter007:
Noticed that questions could not be edited due to a ForeignKey constraint. Removed that constraint.

Also, noticed that templates could not be saved either as `Drafts` or `Published` due to some updates that were needed related to the `isDirty` field.

PR #211, opened 2025-02-22
[Bugs] - Fix Template Builder Flow bugs

jupiter007:
## Description

Need to fix some issues in the Template Builder flow.

- Questions could not be saved due to a ForeignKey Constraint that was removed
- Templates could not be `Saved as Draft` or `Published` due to `isDirty` being set to `FALSE`. So, whenever a Section or Question is mutated, we should update the flag to `TRUE` so that the user can save.
-- When `Save as Draft` is selected, we still need the template to be marked as `isDirty`, otherwise, the `Publish` request will not go through.

Fixes # ([210](https://github.com/CDLUC3/dmsp_backend_prototype/issues/210))

## Type of change

## How Has This Been Tested?
Tested manually, and also updated unit tests.

## Checklist:


data-migrations/2024-09-25-1120-create-questions.sql
```diff
@@ -17,7 +17,6 @@ CREATE TABLE `questions` (
   `modified` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   FOREIGN KEY (templateId) REFERENCES templates(id) ON DELETE CASCADE,
   FOREIGN KEY (sectionId) REFERENCES sections(id) ON DELETE CASCADE,
-  FOREIGN KEY (questionTypeId) REFERENCES questionTypes(id),
```
briri, line 20: This shouldn't be removed. Are we saving questions with a null here for some reason?

src/utils/templateUtils.ts
briri, line 1: Can we move this code onto the template model?

briri (changes requested):

data-migrations/2024-09-25-1120-create-questions.sql
jupiter007: Thanks @briri. I thought it was preventing me from updating questions, but realized that it was because the frontend was converting questionTypeId to zero when it was `null`. I added it back.

briri (approved):
Thanks for making those changes

Issue #215, opened 2025-02-26
Fix a bug with parentResearchDomain always returning null


jupiter007:
While working on https://github.com/CDLUC3/dmsp_frontend_prototype/issues/351, I noticed that the backend always returns null for parentResearchDomain.

PR #216, opened 2025-02-26
[Bug] - Update researchDomain because parentResearchDomain is always returning null

jupiter007:
## Description

While I was working on [#351](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/351), I noticed that I wasn't able to get any data returned for parentResearchDomain. It was always null, even though a parentResearchDomainId existed.

Fixes # ([215](https://github.com/CDLUC3/dmsp_backend_prototype/issues/215))

## Type of change

## How Has This Been Tested?
Manually tested

## Checklist:

### DIFF
--- a/src/controllers/signinController.ts
+++ b/src/controllers/signinController.ts
@@ -17,10 +17,6 @@ export const signinController = async (req: Request, res: Response) => {
     if (user) {
       const { accessToken, refreshToken } = await generateAuthTokens(context, user);
 
-console.log('User:', user);
-console.log('Access Token:');
-console.log(accessToken);
-
       // Record the login and generate the tokens
       if (accessToken && refreshToken) {
         // Set the tokens as HTTP only cookies
--- a/src/models/ResearchDomain.ts
+++ b/src/models/ResearchDomain.ts
@@ -9,6 +9,7 @@ export class ResearchDomain extends MySqlModel {
   public name: string;
   public uri: string;
   public description?: string;
+  public parentResearchDomainId?: number;
   public parentResearchDomain: ResearchDomain;
 
   private tableName = 'researchDomains';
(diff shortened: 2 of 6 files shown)


src/controllers/signinController.ts
```diff
@@ -17,10 +17,6 @@ export const signinController = async (req: Request, res: Response) => {
     if (user) {
       const { accessToken, refreshToken } = await generateAuthTokens(context, user);
 
-console.log('User:', user);
```
jupiter007, line 20: just removed these vestigial console logs

src/resolvers/project.ts
```diff
@@ -179,4 +182,13 @@ export const resolvers: Resolvers = {
       );
     },
   },
+  ResearchDomain: {
```
briri, line 185: This looks like a duplicate of what was added to the ResearchDomain resolver. Is it actually needed here on the project resolver?

src/models/ResearchDomain.ts
```diff
@@ -20,6 +21,7 @@ export class ResearchDomain extends MySqlModel {
     this.name = options.name;
     this.uri = options.uri;
     this.description = options.description;
+    this.parentResearchDomainId = options.parentResearchDomainId;
```
briri, line 24: good catch!

briri (changes requested):
just one comment about the duplicate resolver for ResearchDomain

src/resolvers/project.ts
jupiter007: Yes thanks @briri. I removed it. Thanks.

src/models/ResearchDomain.ts
jupiter007: Thanks for your review @briri

briri (approved):
LGTM. Thanks for making the change @jupiter007

PR #227, opened 2025-03-07
minor tweak to migration file because mysql doesn't allow JSON fields…

briri:
Discovered a bug when trying to rebuild the database from scratch in the AWS dev environment.

MySQL doesn't allow JSON columns to have a default value.


jupiter007 (approved):

jupiter007:
LGTM

Issue #223, opened 2025-03-05
Update publishedTemplates resolver to optimize query


briri:
We need to add a PublishedTemplateSearchResult to the schema and update the resolver to run a query using joins instead of chaining resolvers.

A corresponding PR will need to be made to the frontend that updates the calls to publishedTemplates

PR #234, opened 2025-03-19
Optimized template, versionedTemplate and Project queries

### LINKED ISSUE #218: Optimize the template queries
The template pages are firing off hundreds of chained resolver queries.

It would be better to write an optimized query with joins that returns everything needed at once

briri:
## Description

Fixes #218 and #223

- Added `TemplateSearchResult`, `VersionedTemplateSearchResult` and `ProjectSearchResult` classes to their corresponding models.
- Updated `Template`, `VersionedTemplate` and `Project` schemas and resolvers to use the new search result objects
- Fixed a bug with the `projectCollaborators` table definition which had an FKey on the `plans` table instead of `projects`

**Note that data migrations need to be run to apply the DB change**

## Type of change

## How Has This Been Tested?

- Added unit tests for new objects
- Verified the queries were working in the Apollo explorer
- Verified that the UI changes in PR are also still working using these changes

## Checklist:

### DIFF
--- a/src/models/Project.ts
+++ b/src/models/Project.ts
@@ -1,6 +1,124 @@
 import { MyContext } from "../context";
 import { validateDate } from "../utils/helpers";
 import { MySqlModel } from "./MySqlModel";
+export class ProjectSearchResult {
+  public id: number;
+  public title: string;
+  public abstractText?: string;
+  public startDate?: string;
+  public endDate?: string;
+  public researchDomain?: string;
+  public isTestProject: boolean;
+  public created: string;
+  public createdById: number;
+  public createdByName: string;
+  public modified: string;
+  public modifiedById: number;
+  public modifiedByName: string;
+  public collaborators: { name: string, accessLevel: string, orcid: string }[];
+  public contributors: { name: string, role: string, orcid: string }[];
+  public funders: { name: string, grantId: string }[];
+
+  constructor(options) {
+    this.id = options.id;
+    this.title = options.title;
+    this.abstractText = options.abstractText;
+    this.startDate = options.startDate;
+    this.endDate = options.endDate;
+    this.researchDomain = options.researchDomain;
+    this.isTestProject = options.isTestProject;
+    this.created = options.created;
+    this.createdById = options.createdById;
+    this.createdByName = options.createdByName;
+    this.modified = options.modified;
+    this.modifiedById = options.modifiedById;
+    this.modifiedByName = options.modifiedByName;
+    this.collaborators = options.collaborators;
+    this.contributors = options.contributors;
+    this.funders = options.funders;
+  }
+
+  // Return all of the projects for the User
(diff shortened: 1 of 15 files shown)


src/models/Project.ts
```diff
@@ -1,6 +1,124 @@
...
+    this.modifiedById = options.modifiedById;
+    this.modifiedByName = options.modifiedByName;
+    this.collaborators = options.collaborators;
+    this.contributors = options.contributors;
+    this.funders = options.funders;
+  }
+
+  // Return all of the projects for the User
+  static async findByUserId(reference: string, context: MyContext, userId: number): Promise<ProjectSearchResult[]> {
+    const sql = 'SELECT p.id, p.title, p.abstractText, p.startDate, p.endDate, p.isTestProject, ' +
```
jupiter007, line 43: I ran this with the corresponding frontend branch, and when I query `MyProjects` I get this error 
```
Error","message":"Column 'projectId' in SELECT is ambiguous","stack":"Error: Column 'projectId' in SELECT is ambiguous\n 
```

jupiter007 (changes requested):
Just double check that there are no errors when testing out the updated resolvers.

src/models/Template.ts
```diff
@@ -7,6 +7,65 @@ export enum TemplateVisibility {
...
+  static async findByAffiliationId(
+    reference: string,
+    context: MyContext,
+    affiliationId: string
+  ): Promise<TemplateSearchResult[]> {
+    const sql = 'SELECT t.id, t.name, t.description, t.visibility, t.bestPractice, t.isDirty, ' +
+                  't.latestPublishVersion, t.latestPublishDate, t.ownerId, a.displayName, ' +
+                  't.createdById, TRIM(CONCAT(cu.givenName, CONCAT(\' \', cu.surName))) as createdByName, t.created, ' +
+                  't.modifiedById, TRIM(CONCAT(mu.givenName, CONCAT(\' \', mu.surName))) as modifiedByName, t.modified ' +
+                'FROM templates t ' +
```
jupiter007, line 58: I noticed that when I log in as `super@example.com`, I was no longer getting any results for the myTemplates query. It turns out that's because of the `'INNER JOIN affiliations a ON a.uri = t.ownerId ' +` line in the query. I'm not sure what user is best to log in with, but it would be nice to add an affiliation in the data-migrations for the `super@example.com` user's affiliation if people will be using that one for testing.

jupiter007 (approved):
LGTM. I think I hadn't approved yet because I noticed that templates were no longer coming up for `super@example.com` when running `myTemplates` query due to a new inner join. 
But I guess we can update the data later.

briri:
Ok. Yeah, if the super admin doesn't "own" any templates then they shouldn't probably show up with that query. We can use the other endpoints (or add new ones) for super admins, but lets wait to discuss/figure out what the workflow would look like for a super admin needing to update another user's template.

PR #237, opened 2025-03-20
External Project Search and Import

jdddog:
## Description

External project search and project import.

## Type of change

## How Has This Been Tested?
I have added unit tests for some parts of the code and also tested the new query and mutation locally.

## Checklist:


src/models/Affiliation.ts
```diff
@@ -106,7 +105,7 @@ export class Affiliation extends MySqlModel {
     this.feedbackEmails = options.feedbackEmails;
 
     // We're proxying calls to funder APIs through the DMPHub API for now. This may change in the future
-    this.apiTarget = prepareAPITarget(options.apiTarget);
+    this.apiTarget = options.apiTarget; // prepareAPITarget(options.apiTarget);
```
jdddog, line 108: The .get method from DMPHubAPI already sets a base URL, so I thought that it made it easier to not add the DMPHub URL to apiTarget, but I'm not sure if this breaks how you are planning to use this for other things? So I could also remove the URL inside getAwards or before calling getAwards.

src/resolvers/project.ts
```diff
@@ -151,6 +268,58 @@ export const resolvers: Resolvers = {
...
+              await toCreate.create(context, projectId);
+            }
+
+            // Update project contributors
+            for (const contrib of input.contributors) {
+              // Add project contributor
+              const toCreate = new ProjectContributor(contrib);
+              await toCreate.create(context, projectId);
+
+              // TODO: raise errors
```
jdddog, line 304: Adding project funders and contributors can fail, so do I need to combine all of the errors and return them as a part of the response? At the moment I'm returning the project object, which will only have its own errors.

src/resolvers/project.ts
```diff
@@ -48,6 +58,113 @@ export const resolvers: Resolvers = {
...
+
+      // Normalises dates
+      const normaliseDate = (date: string | null): string => {
+        if (date === null || date === undefined) {
+          return null;
+        }
+
+        // Split into parts, convert to integers and only continue if 3 integer parts
+        const parts = date
+          .split("-")
```
jupiter007, line 73: Will the dates always be expected to use hyphens?

src/resolvers/project.ts
```diff
@@ -48,6 +58,113 @@ export const resolvers: Resolvers = {
...
+          .filter((p) => Number.isInteger(p));
+        if (parts.length !== 3) {
+          return null;
+        }
+
+        // Convert to numbers
+        return `${parts[0]}-${String(parts[1]).padStart(2, "0")}-${String(parts[2]).padStart(2, "0")}`;
+      };
+
+      const normaliseORCID = (input: string | null) => {
```
jupiter007, line 84: Would it make sense to move some of these helper functions out to utils?

src/schemas/project.ts
```diff
@@ -1,12 +1,18 @@
...
-export const typeDefs = gql`
+export const projectTypeDefs = gql`
   extend type Query {
     "Get all of the user's projects"
     myProjects: [Project]
 
     "Get a specific project"
     project(projectId: Int!): Project
+
+    "Search for external projects"
```
jupiter007, line 14: L14: It might be helpful to provide more info on what the external data sources are.

src/services/userService.ts
```diff
@@ -55,7 +55,7 @@ export const anonymizeUser = async (context: MyContext, user: User): Promise<Use
 
   // Anonymize the user's information
   user.email = `${randomHex(6)}@deleted-account.${generalConfig.domain}`;
-  user.password = await user.hashPassword(generateRandomPassword()),
+  user.password = await user.hashPassword(generateRandomPassword());
```
jupiter007, line 58: nice catch

src/schemas/user.ts
```diff
@@ -178,7 +178,7 @@ export const typeDefs = gql`
     languageId: String
   }
 
-  input updateUserNotificationsInput {
+  input UpdateUserNotificationsInput {
```
jupiter007, line 181: Thanks for updating these to be Pascal case

src/resolvers/project.ts
```diff
@@ -48,6 +58,113 @@ export const resolvers: Resolvers = {
...
+        if (input === null || input === undefined) {
+          return null;
+        }
+        return input.trim().replace(/^https?:\/\/orcid\.org\//, '');
+      }
+
+      // Parses a contributor into a standard format
+      const parseContributor = (
```
briri, line 92: Please move this function into the `src/services/commonStandardService.ts`. I've got logic there that convert from our MySQL DB records into the JSON common standard format. I'm going to want to eventually have have functions to go in the other direction so this will be a good start

src/resolvers/project.ts
briri: Yes, we will want to capture and report those errors as well. I've done something similar here: https://github.com/CDLUC3/dmsp_backend_prototype/blob/864312ac43052685baaf81c767133d323708e9a0/src/resolvers/contributor.ts#L103
You can combine all of the errors and add them to `project.addError('general', 'the errors messages')`

src/schemas/user.ts
```diff
@@ -12,9 +12,9 @@ export const typeDefs = gql`
 
   extend type Mutation {
     "Update the current user's information"
-    updateUserProfile(input: updateUserProfileInput!): User
+    updateUserProfile(input: UpdateUserProfileInput!): User
```
briri, line 15: thanks for capitalizing these!

src/resolvers/project.ts
jupiter007: > Is it possible @jupiter007 to ensure that the dates the frontend sends will always come through with hyphens instead of slashes? This will reduce confusion with locales

Yes I will make a note of that

src/resolvers/project.ts
jdddog: Thanks @briri, I'll do that.

src/datasources/dmphubAPI.ts
```diff
@@ -222,6 +222,65 @@ export class DMPHubAPI extends RESTDataSource {
       throw(err);
     }
   }
+
+  // Search for awards via the DMPHub
```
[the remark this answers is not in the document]
jdddog: Thanks @jupiter007, I can add JSDocs here.

src/datasources/dmphubAPI.ts
```diff
@@ -222,6 +222,65 @@ export class DMPHubAPI extends RESTDataSource {
...
+      let path = apiTarget;
+      const queryString = params.toString();
+      if(queryString){
+        path += `?${queryString}`;
+      }
+
+      formatLogMessage(context).debug(`${reference} Calling DMPHub getAwards: ${this.baseURL}/${path}`)
+      const response = await this.get(path);
+      if (response?.status === 200 && Array.isArray(response?.items)) {
+        console.log(JSON.stringify(response.items));
```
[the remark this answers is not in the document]
jdddog: I left this in by mistake, I'll change it to use formatLogMessage

src/resolvers/project.ts
jdddog: These are the dates that are returned by the DMP Hub API, which come from the Crossref Metadata, NIH and NSF endpoints. They all use hyphens, but the month and day for the Crossref Metadata based dates don't have leading zeros, whereas the others do, which is because the Crossref dates are constructed from a list of integers, e.g. `date-parts: [2021,1,1]`.

The leading zeros could also be added here: https://github.com/CDLUC3/dmsp_api_prototype/blob/main/lambdas/api/get_awards_crossref/app.rb#L257

I will move it to `src/utils/helper.ts`.

src/schemas/project.ts
jdddog: Yeah sure

src/resolvers/project.ts
jdddog: I wondered about using `stripIdentifierBaseURL`, at the moment it doesn't handle the case where the ORCID URL starts with `http://` rather than `https://`, and some of the ORCID URLs start with `http://`.

I could modify it to also handle this case? 

Do you use generalConfig.orcidBaseURL, orcidBaseURL.rorBaseURL etc for other things?

src/resolvers/project.ts
```diff
@@ -48,6 +58,113 @@ export const resolvers: Resolvers = {
...
+        }
+
+        const nameParts = input.name.split(",").map((t) => t.trim());
+        const surName = nameParts[0];
+        const givenName = nameParts[nameParts.length - 1];
+        const email = input.mbox;
+        let orcid = null;
+
+        // ORCID ID
+        // TODO: update stripIdentifierBaseURL to handle http for ORCID as well?
```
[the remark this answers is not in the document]
jdddog: Yeah sure, will do.

src/resolvers/project.ts
jdddog: Yeah sure, will do.

jdddog:
Thanks @jupiter007 and @briri for your reviews, that's really helpful, I'll make those changes.

src/datasources/dmphubAPI.ts
```diff
@@ -222,6 +222,65 @@ export class DMPHubAPI extends RESTDataSource {
...
+    apiTarget: string,
+    awardId: string | null = null,
+    awardName: string | null = null,
+    awardYear: string | null = null,
+    piNames: string[] | null = null,
+    reference = 'dmphubAPI.getAwards'
+  ): Promise<DMPHubAward[]> {
+    try {
+      // Build query parameters
+      const params = new URLSearchParams();
```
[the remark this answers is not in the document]
jdddog: I'm not sure? I will think about it a bit more.

There are some cases when calling the Crossref Metadata endpoint, where it needs certain parameters otherwise it doesn't return anything (that might also be a bug, I'll need to try it again), so for that case it will do.

src/resolvers/project.ts
briri: yes, you can absolutely update that helper method to handle `http`. If you do, please make sure the other URL types also handle the different in protocol

briri: it doesn't look like those `generalConfig` URLs are used elsewhere except for some old resolver scalars that we ended up not using, so you can go ahead and update them if needed. If you do, there is an `.env.example` file that should be updated and then just make a note in the PR so that we all know to update our own `.env` files

jdddog:
Hey @briri and @jupiter007 .

I've made most of those changes.

Here are some comments about the ORCID related changes:
* Added the ORCID URL prefix before each ORCID ID in `data-migrations/2025-01-06-1144-seed-projects.sql`. Should this instead be added to a new migration?
* Since we are storing the full ORCID URL, we need to change `http://` to `https://` as there are still ORCID URLs in the funder APIs with `http://` and this will cause contributors to not match when they should.
* Consolidated ORCID functionality in `src/models/User.ts`, `src/models/Contributor.ts` and `src/resolvers/scalars/orcid.ts`, moving it to `src/utils/helpers.ts`.
* The ORCID regex here: https://github.com/CDLUC3/dmsp_backend_prototype/blob/b26ba86198751276cc54bbb4687f45c851d4e41a/src/utils/helpers.ts#L7 limits the ORCID domain to `sandbox.orcid.org` or `orcid.org`, which could cause problems when the ORCID base URL can be changed here: `src/__tests__/setup.ts` and in the tests.

Still to do: write tests for `getAwards` function and check that default role is created for contributors.

src/models/User.ts
```diff
@@ -99,7 +84,7 @@ export class User extends MySqlModel {
     if (!supportedLanguages.map((l) => l.id).includes(this.languageId)){
       this.languageId = defaultLanguageId;
     }
-    this.orcid = this.orcid? User.formatORCID(this.orcid) : null;
+    this.orcid = this.orcid? formatORCID(this.orcid) : null;
```
briri, line 87: thanks for consolidating all of these into the new helper function :)

src/resolvers/project.ts
```diff
@@ -151,6 +204,85 @@ export const resolvers: Resolvers = {
...
+            }
+
+            // Update project
+            const toUpdateProject = new Project(input.project);
+            const updatedProject = await toUpdateProject.update(context);
+            if(!updatedProject){
+              toUpdateProject.addError('general', 'Unable to update project');
+              return toUpdateProject;
+            }
+
```
briri, line 231: It's possible for the `toUpdateProject.update` function to return the project with errors (instead of just null). It's probably good to check if it has errors , something like this: 
```
if (updatedProject && !updatedProject.hasErrors()) {
  // ... add the funders
  // ... add the contributors 
}
```

src/schemas/contributor.ts
```diff
@@ -16,7 +16,7 @@ export const typeDefs = gql`
     "Add a contributor to a research project"
     addProjectContributor(input: AddProjectContributorInput!): ProjectContributor
     "Update a contributor on the research project"
-    updateProjectContributor(input: updateProjectContributorInput!): ProjectContributor
+    updateProjectContributor(input: UpdateProjectContributorInput!): ProjectContributor
```
briri, line 19: good catch!

src/utils/helpers.ts
```diff
@@ -118,4 +147,38 @@ export function getCurrentDate(): string {
 // Generate a random hex code
 export function randomHex(size: number): string {
   return [...Array(size)].map(() => Math.floor(Math.random() * 16).toString(16)).join('');
-}
\ No newline at end of file
+}
+
+// Normalises dates
+export function normaliseDate (date: string | null): string {
```
briri, line 153: Thanks @jdddog these all look good to me

briri (approved):

PR #191, opened 2025-01-31
Add the requestId to the Apollo context and include it in logs

briri:
## Description

Fixes #190 

- Added `requestId` to the Apollo context
- Updated the `formatLogMessage` function so that it includes the following fields (when available) on each request: `requestId`, `jti` and `userId`

@jupiter007 the core changes are in `src/context.ts` and `src/logger.ts` the majority of the other changes are updated to accommodate the fact that `formatLogMessage` now take the Apollo context as its argument instead of logger

## Type of change
Please delete options that are not relevant

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Include any relevant details for your test configuration.

All unit tests are passing, the Apollo server loads and I can run queries in the explorer

## Checklist:
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I updated the CHANGELOG.md and added documentation if necessary
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #256, opened 2025-04-21
Feature/br popular funders

briri:
## Description

Add a new query resolver to fetch the top 20 Funders. The query determines funder popularity based on the number of non-Test projects that were created over the past year that are tied to the funder.

@andrewebdev you will be able to use this on the Funder search page using the following query:
```
query PopularFunders {
  popularFunders {
    id
    uri
    displayName
    nbrPlans
  }
}
```

Fixes [#191](https://github.com/CDLUC3/dmsp_aws_prototype/issues/191)

## Type of change

## How Has This Been Tested?

Added unit tests. Also verified that the query returns results in the Apollo Server explorer

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ### Added
+- Added `popularFunders` query to the `affiliations` resolver
 - Added `paginateResults` helper function to `MySQLModel`.
 - Added `accessLevel` to projectCollaborator and removed `userId`
 - Added new resolvers related to `projectCollaborators`. Also, when project is created, automatically add user as `projectCollaborator` with `access level`= `OWN`


briri:
experimenting with the CoPilot PR review process. It caught a couple of typos

jupiter007:
> experimenting with the CoPilot PR review process. It caught a couple of typos

nice

src/models/Affiliation.ts
```diff
@@ -351,3 +351,41 @@ export class AffiliationSearch {
...
+    lastYear.setFullYear(today.getFullYear() - 1);
+    const startDate = lastYear.toISOString().split('T')[0];
+    const endDate = today.toISOString().split('T')[0];
+
+    // Get the top 20 funders based on the number of plans created in the past year
+    const sql = 'SELECT a.id, a.uri, a.displayName, COUNT(p.id) AS nbrPlans ' +
+                'FROM affiliations a ' +
+                'LEFT JOIN projectFunders pf ON pf.affiliationId = a.uri ' +
+                'LEFT JOIN projects p ON p.id = pf.projectId ' +
+                'WHERE a.active = 1 AND a.funder = 1 AND p.isTestProject = 0 AND p.created BETWEEN ? AND ? ' +
```
jupiter007, line 382: Would we want to factor in status at all? For example, would we want to count those projectFunders that were DENIED?

jupiter007 (approved):
@briri. LGTM. 

Remember to update your CHANGELOG.md file.

Also, I just had one question about whether we need to factor in projectFunder `status` at all in the query. Probably not, but thought I would check.

src/models/Affiliation.ts
briri: No I don't think so. The status indicates whether or not a grant was received. We just want to know which ones have popular templates.

briri:
updated CHANGELOG

PR #267, opened 2025-04-30
Feature/br pagination

### LINKED ISSUE #180: Allow pagination for management search results (users, plans, orgs, templates)
Many of the tables we'll create for admin management can have tens or even hundreds of thousands of results, like a super admin looking at a table of all users.  We need to set up the backend to do pagination so we're not pulling all of those at once.

briri:
## Description

Fixes #180

**There is a corresponding PR in the frontend [#440](https://github.com/CDLUC3/dmsp_frontend_prototype/pull/440) to accommodate these changes!**

Support for pagination on `affiliations`, `publishedTemplates`, `users`, `myTemplates`, `repositories`, `myProjects`, `metadataStandards`, `licenses`, `affiliations` queries.

We now support **cursor** pagination which should be used for endless-scroll / "Load more" scenarios and **offset** pagination used for more traditional "first, previous, next, last" style navigation.

**Please note that these changes will break query functionality in the frontend.** The queries listed above have changed!. They now return pagination metadata along with the results. 

**Cursor based pagination:** 
This is the default. If you specify no pagination options it will be used. This query response looks like:
```
{
  "totalCount": 35,
  "limit": 20,
  "hasNextPage": true,
  "cursor": "unique_value3",
  "items": [ 
    "... the list of items (e.g. Affiliations)" 
  ]
}
```
The same query can then be called by specifying `{ "cursor": "unique_value3" }` to get the next page of results.

If you need/want to use offset pagination, the initial query should include `{ "offset": 0 }` which will return the first page of results that looks something like this:
```
{
  "totalCount": 35,
  "limit": 20,
  "currentOffset": 0,
  "hasNextPage": true,
  "hasPreviousPage": false,
  "availableSortFields: ["t.name", "t.modified"]
  "items": [ 
    "... the list of items (e.g. Affiliations)" 
  ]
}
```
The same query can then be called by specifying `{ "offsets": 20 }`. You can perform calculations using the totalCount, limit and currentOffset to determine individual page offsets.

You can use the fields in `availableSortFields` to specify sort options when using Offset pagination `{ "sortField": "t.modified", "sortDir": "DESC" }`

Note the default `limit` is 20 with a max of 100. You can specify an exct number in your query.

- Added new `queryWithPagination` function to `src/models/MySqlModel.ts`
- Added new schema types `PaginationOptions` and `PaginatedQueryResults`
- Updated aforementioned resolvers and models to use the new schema types and `MySqlModel` functions
- Added tests for new functions and updated existing tests that were broken by changes

## Type of change

## How Has This Been Tested?
- Unit tests
- Affiliations resolver test
- Manual Apollo explorer queries

## Checklist:


briri:
@fraserclark @jupiter007 and @andrewebdev updating the Sort info in the response looks to be complicated enough to warrant a separate ticket. I created [this one](https://github.com/CDLUC3/dmsp_backend_prototype/issues/268) to track the work and the discussion. Please take a look at the proposed format and let me know what you think.

src/models/License.ts
```diff
@@ -124,13 +126,52 @@ export class License extends MySqlModel {
...
+    name: string,
+    options: PaginationOptions = License.getDefaultPaginationOptions(),
+  ): Promise<PaginatedQueryResults<License>> {
+    const whereFilters = [];
+    const values = [];
+
+    // Handle the incoming search term
+    const searchTerm = (name ?? '').toLowerCase().trim();
+    if (searchTerm) {
+      whereFilters.push('(LOWER(l.name) LIKE ? OR LOWER(l.description) LIKE ?)');
```
jupiter007, line 141: L141: Just a heads up that I read that using `LOWER` in your SQL can degrade performance.

src/types/general.ts
```diff
@@ -0,0 +1,34 @@
+export type PaginationOptions = PaginationOptionsForOffsets | PaginationOptionsForCursors;
```
jupiter007, line 1: L1: I think this `union` type blocks access to non-shared properties. Typescript will then treat PaginatedOptions as having only the properties that exist on both interfaces. So if you try and access `options.offset` without narrowing the type, TypeScript will raise an error.

You could possibly add a `type' field to both interfaces to help with narrowing.

src/models/Affiliation.ts
```diff
@@ -331,24 +328,56 @@ export class AffiliationSearch {
...
-    if (options.name) {
-      sql += ' AND LOWER(searchName) LIKE ?'
-      vals.push(`%${options.name.toLowerCase()}%`);
+  static async search(
+    reference: string,
+    context: MyContext,
+    name: string,
+    funderOnly: boolean,
+    options: PaginationOptions = Affiliation.getDefaultPaginationOptions(),
+  ): Promise<PaginatedQueryResults<AffiliationSearch>> {
```
jupiter007, line 337: L337: According to the PaginatedQueryResults<AffiliationSearch> type, we should be returning the `availableSortFields`, but are we? It looks like you are defining it on like 363, but it's not attached to the response.

src/models/Affiliation.ts
```diff
@@ -331,24 +328,56 @@ export class AffiliationSearch {
...
+    // Handle the incoming search term
+    const searchTerm = (name ?? '').toLowerCase().trim();
+    if (searchTerm) {
+      whereFilters.push('(LOWER(a.searchName) LIKE ?)');
+      values.push(`%${searchTerm}%`);
     }
-    if (options.funderOnly) {
-      sql += ' AND funder = 1';
+    if (funderOnly) {
+      whereFilters.push('funder = 1');
```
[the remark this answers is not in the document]
briri: yes, I will add that. MySQL is smart enough to figure out that affiliations is the only table with that column but I should add the prefix to be consistent

src/models/Affiliation.ts
```diff
@@ -331,24 +328,56 @@ export class AffiliationSearch {
...
+    // Specify the field we want to use for the count
+    options.countField = 'a.id';
+
+    // if the options are of type PaginationOptionsForOffsets
+    if ('offset' in options && !isNullOrUndefined(options.offset)) {
+      // Specify the fields available for sorting
+      options.availableSortFields = ['a.displayName', 'a.created'];
+    } else if ('cursor' in options) {
+      // Specify the field we want to use for the cursor (should typically match the sort field)
+      options.cursorField = 'LOWER(REPLACE(CONCAT(a.name, a.id), \' \', \'_\'))';
```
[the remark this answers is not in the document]
briri: name is required on the table so it should be fine. The cursor needs tone something unique within the record so that we can do a `>` query.

src/models/License.ts
briri: Do you have an alternative? It's necessary when doing these kinds of searches. We want `mit` to find `MIT`.

I suppose at some later date, we could have an additional column in these tables that stores the lower case version of the text to help optimize searches

jupiter007: I think the issue is a combination of using `LOWER` and `LIKE` that may cause performance issues, especially when dealing with large datasets.

According to some reading I was doing, using `LIKE` has to search `name` or `description` character by character. I'm not sure what the best solution is though. I read that there is something called `Full Text search` that can speed things up.

briri:
Thanks for the review @jupiter007. I have made the suggested changes. Adding a `type` to the `PaginationByX` interfaces required a bit of work but I like that it is cleaner.

jupiter007:
> Thanks for the review @jupiter007. I have made the suggested changes. Adding a `type` to the `PaginationByX` interfaces required a bit of work but I like that it is cleaner.

Thanks for making the change. Yea, type unions have foiled me in the past. 

In regards to using `LOWER` and `LIKE` in our queries, we might want to revisit that in the future if there are performance issues with larger datasets.

jupiter007 (approved):

Issue #271, opened 2025-05-05
Bug -


jupiter007:
Discovered some bugs while working on the frontend ticket: [265](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/265).

Need to set bestPractice to be required in the `updateTemplate` schema, since its non-nullable in the Template type, and was preventing the `updateTemplate` mutation from successfully completing without it. Also, need to add an update to `createTemplateVersion`, so that errors from the `generateTemplateVersion` will be caught and passed back in graphql response. Previously, when trying to save a DRAFT of a template, the mutation wouldn't return an error to the client, even though the `Save draft` did not successfully complete

PR #272, opened 2025-05-05
Addressed small bugs when trying to `updateTemplate` and `createTemplateVersion`

jupiter007:
## Description

Set bestPractice to be required in the `updateTemplate` schema, since its non-nullable in the Template type, and was preventing any updates of the template. Also, added an update to `createTemplateVersion`, so that errors from the `generateTemplateVersion` will be caught and passed back in graphql response. Previously, when trying to save a DRAFT of a template, the mutation wouldn't return an error to the client, even though the `Save draft` did not successfully complete.

Fixes # ([271](https://github.com/CDLUC3/dmsp_backend_prototype/issues/271))

## Type of change

## How Has This Been Tested?
Manually tested with client to check that `updateTemplate` and `createTemplateVersion` were successful after change.

## Checklist:

### DIFF
--- a/src/schemas/template.ts
+++ b/src/schemas/template.ts
@@ -12,7 +12,7 @@ export const typeDefs = gql`
     "Create a new Template. Leave the 'copyFromTemplateId' blank to create a new template from scratch"
     addTemplate(name: String!, copyFromTemplateId: Int): Template
     "Update a Template"
-    updateTemplate(templateId: Int!, name: String!, visibility: TemplateVisibility!, bestPractice: Boolean): Template
+    updateTemplate(templateId: Int!, name: String!, visibility: TemplateVisibility!, bestPractice: Boolean!): Template
     "Archive a Template (unpublishes any associated PublishedTemplate"
     archiveTemplate(templateId: Int!): Template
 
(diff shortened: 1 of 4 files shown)


src/schemas/template.ts
```diff
@@ -12,7 +12,7 @@ export const typeDefs = gql`
     "Create a new Template. Leave the 'copyFromTemplateId' blank to create a new template from scratch"
     addTemplate(name: String!, copyFromTemplateId: Int): Template
     "Update a Template"
-    updateTemplate(templateId: Int!, name: String!, visibility: TemplateVisibility!, bestPractice: Boolean): Template
+    updateTemplate(templateId: Int!, name: String!, visibility: TemplateVisibility!, bestPractice: Boolean!): Template
```
briri, line 15: I think it would be better to just update [this line](https://github.com/CDLUC3/dmsp_backend_prototype/blob/b43496f55b494f1aae203713a4038432d7000beb/src/resolvers/template.ts#L151) instead to ensure that the value for `bestPractice` is set. Only a Super Admin should be able to set this.

jupiter007:
> fine with this solution for now, or if you have time and want to try to see if you can determine why it's not defaulting that field properly that would be good too.
> 
> If you go with this as is, can you please create a ticket (in the backlog) to investigate why its not defaulting? I suspect it may be a larger issue and impacting other places as well.

@briri , it turned out that I just needed to fix the line where `templateInstance.bestPractice` is set in `updateTemplate`. Essentially if `bestPractice` was undefined, `templateInstance.bestPractice` would be set to `undefined` rather than using the existing `bestPractice` value. I added code to fallback to the existing value if no `bestPractice` value is passed.

PR #284, opened 2025-05-20
Convert questions to store the type info as JSON

### LINKED ISSUE #253: Update questions table and model to use JSON
Switch the questions.questionText field from TEXT to JSON

- Drop the existing questionOptions table (and model, resolver, schema)
- Convert the question details into the new JSON format (see below)
- Update the existing Question model so that it handles TextField and TextArea question types
- Add new classes that extend the base Question model that handle the other question types
- We could decide whether its best to just pass the JSON as-is to the frontend to let it convert that into UI elements or we could do so in the models on the backend

We can reference  https://jsonforms.io/ for some ideas on structure. Create most complex question type structure first, and work backwards.

Examples of the way the JSON could be structured:

For text based fields like text field, text area, url, etc. We could store it like this:

text: "This is my question"
For multiple choice fields like checkboxes, select boxes, etc, we store the details about the range in a format like this:

```
{ 
  text: "What is your favorite color?", 
  options: [{ label: "Blue" }, { label: "Green", isDefault: true }] 
}
```
For a data range, the questions.questionTypeId could equal something like "DATE_RANGE" and we store the details about the range in a format like this:

```
{
  text: "For what period will you collect your data?",
  fields: [
    { label: "Start month", type: "DATE" },
    { label: "End month", type: "DATE" },
  ]
}
```
For research outputs, the questions.questionTypeId could equal something like "TABLE" and we store the details about the table in a format like this:

{
  text: "Please list your intended datasets",
  fields: [
    { label: "Title", type: "TEXT" },
    { label: "Description", type: "TEXT" },
    { label: "May contain PII?", type: "CHECKBOX" },
    { label: "Release date", type: "DATE" },
    { label: "Repositories", type: "REPO_SELECTION" },
  ]
}
This should give us the flexibility we need and also support future enhancements that allow the admin to define their own tables or add custom columns to things like the research output question type.

briri:
## Description

Fixes #253

Moved the QuestionType and QuestionOption information into the new `questionType` JSON fields.

- Add JSON column `json` to `questionTypes`, `questions`, `versionedQuestions` and `answers` tables
- Added script to populate the new columns
- Dropped `questionOptions` table
- Dropped the `answerText` field from `answers`
- Removed the old `QuestionOptions` schema and resolvers
- Added the new [@dmptool/types](https://github.com/CDLUC3/dmptool-types) package 
- Updated the `QuestionType`, `Question` and `VersionedQuestion` schemas to allow `json: String` to be used in GraphQL query
- Updated `Question` model to access `questionType` input, parse the JSON using the Zod schemas provided by `@dmptool/types`, and populate the `json` with the parsed JSON.
- Removed the `getQuestionOptionsToRemove` from `questionService`.

## Type of change

## How Has This Been Tested?

- Added new Unit tests and ensured existing tests pass
- Verified queries and mutations for questions via Apollo explorer

## Checklist:


data-migrations/2025-05-13-1528-addQuestionJSON.sql
```diff
@@ -0,0 +1,10 @@
+# Add the questionJSON column to the questions table
```
jupiter007, line 1: L1 - Do you mean `questionType` and not `questionJSON` here?

data-migrations/2025-05-14-0807-seed-questionJSON.sql
```diff
@@ -0,0 +1,68 @@
+
+# Set the questionType field for 'Text Field' questions
+UPDATE `questions` SET `questionType` = '{"type":"text","meta":{"schemaVersion":"1.0"}}'
+WHERE `questionTypeId` IN (SELECT `id` FROM `questionTypes` WHERE `name` = 'Text Field');
+
+UPDATE `versionedQuestions` SET `questionType` = '{"type":"text","meta":{"schemaVersion":"1.0"}}'
+WHERE `questionTypeId` IN (SELECT `id` FROM `questionTypes` WHERE `name` = 'Text Field');
+
+UPDATE `answers` SET `json` = CONCAT('{"type":"text","answer":"', `answers`.`answerText`, '","meta":{"schemaVersion":"1.0"}}')
```
jupiter007, line 9: I didn't see any data in the `answers` table after the data-migrations.

src/models/Question.ts
```diff
@@ -34,17 +37,37 @@ export class Question extends MySqlModel {
     this.required = options.required ?? false;
     this.displayOrder = options.displayOrder;
     this.isDirty = options.isDirty ?? false;
-    this.questionOptions = options.questionOptions;
+
+    // If questionJSON is not null or undefined and the type is in the schema map
+    if (!isNullOrUndefined(this.questionJSON)) {
+      try {
+        const parsedJSON = JSON.parse(this.questionJSON);
+        if (Object.keys(QuestionSchemaMap).includes(parsedJSON['type'])) {
```
jupiter007, line 45: L45: You could possibly write this so that you don't have to use `Object.keys` here:
```
    // Check if the type exists in the schema map
    const schema = QuestionSchemaMap[parsedJSON['type']];
    if (schema) {
      // Validate and set the questionType
      this.questionType = schema.parse(parsedJSON);
    } else {
      this.addError('questionJSON', `Unknown question type "${parsedJSON['type']}"`);
    }
```

src/models/VersionedQuestion.ts
```diff
@@ -21,24 +24,32 @@ export class VersionedQuestion extends MySqlModel {
...
     this.guidanceText = options.guidanceText;
     this.sampleText = options.sampleText;
     this.required = options.required ?? false;
     this.displayOrder = options.displayOrder;
+
+    // stringify the questionType and set it to the questionJSON
+    try {
+      this.questionJSON = JSON.stringify(this.questionType);
+    } catch (e) {
+      // Add the Zod schema error to the errors object
```
jupiter007, line 39: L39: I think if an error is caught, it's just coming from `JSON.stringify()` and not `Zod` schema, right? I'm not sure you really need a try-catch here either.

jupiter007 (approved):
Hi Brian. Looks good. I just had a couple of comments.

briri:
@jupiter007 I added tests for all of the default question JSON and a data migration that will fix the ones that were invalid. You will need to pull down the latest and run `data-migrations/process.sh`

jupiter007:
> @jupiter007 I added tests for all of the default question JSON and a data migration that will fix the ones that were invalid. You will need to pull down the latest and run `data-migrations/process.sh`

Thanks @briri. I'll take a look.

Issue #278, opened 2025-05-09
Bug - Can no longer update User Profile


jupiter007:
## How to reproduce
1. Log in as super@example.com
2. Navigate to `/account/profile` and try and update the user's first name and last name
3. There are no errors returned, so it says the update was successful
4. But the database is not updated:
-- `updateUserProfile` is trying to call `update` here: `const updated = await new User(user).update(context);`, but it's not passing validation because there is no `createdById` or `modifiedById`.

When we seed our user data, we should populate that data for testing.

bofstein:
Confirmed on dev with super@example, not sure if we need to check with other migrated users

PR #294, opened 2025-05-28
Update profile was not working due to missing createdById and modifiedById values in db

jupiter007:
## Description

- Added a data-migration script to populate the `createdById` and `modifiedById` fields so that `updateProfile` mutation will not break.

Fixes # ([278](https://github.com/CDLUC3/dmsp_backend_prototype/issues/278))

## Type of change

## How Has This Been Tested?
Tested manually and with unit test.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,7 @@
 
 ### Fixed
 - Update profile was not working due to missing `createdById` and `modifiedById` values in db. Added data migration script to populate those fields [#278]
+- Fixed myTemplates query so that `TemplateSearchResult` returns the `ownerDisplayName` specified in schema.
 - Fixed an issue where adding `templateCollaborators` was failing due to the fact that the `userId` field was required.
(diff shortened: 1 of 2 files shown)


data-migrations/2025-05-28-1158-add-createdById-modifiedById-values-to-users.sql
andrewebdev, line 1: This feels a bit strange to have a migration that only targets certain users. I'm not too familiar with the setup and structure here, so I'm just wondering if what's happening here is safe, and that we are not missing some users with this kind of migration where the emails are hard-coded directly into the migration?

andrewebdev (commented):
While the migration looks mostly fine. I'm wondering why emails are hard-coded directly into the migration. If it's okay for this migration, have the email's been double-checked for typos, and are we sure we are assigned correct modified and create ids to the correct users?

andrewebdev (approved):

Issue #200, opened 2025-02-07
Add resolver tests for remaining resolvers starting with project and its associated objects


briri:
We need to go back and add integration tests for all of the resolvers. Start with the project/plans since this is the next logical work area.

We can create separate tickets to go back to add them for templates and users if necessary

bofstein:
@briri to make tickets for what is left and put into appropriate epics

briri:
Added follow up tickets for remaining resolvers
- #325 tests for lookup tables (repositries, licenses, etc.)
- #326 tests for project resolver
- #327 tests for template resolver

briri:
Local test DBs have proven to be unreliable at best and a bit wonky to manage. I have left the branch, `feature/refactor-resolvers` I was working in the repo. We can revisit someday in the future.

It might even prove better to have a completely independent app that hits the AWS development environment.

PR #309, opened 2025-06-13
Refactor the mysql datasource and the way we pass the connection pool around

briri:
## Description

1st part of #200 

**NOTE: this involves some new .env variable! AND run `data-migrations/process.sh`**

- Add an addiitional MySQL test DB to the docker-compose stack
- Updated `process.sh` and `nuke-db.sh` scripts to work with the test DB as well
- Added new ENV variables to the `.env-example`. The docker-compose has the test DB info, but you need the ENV variables in your local `.env` to get the jest resolver tests to work with the new test DB.

Discovered that we weren't properly working with a MySQL connection pool and were instead relying on a single connection. This might have been causing some of the issues we were seeing with too many queries a while back before we optimized some of the search queries.

- Updated the `src/datasources/mysql.ts` to use a pool

Discovered that we were not properly passing the logger, cache and datasources around properly. Also discovered how to pass them through to the non-graphql endpoints.

- Updated the `src/context.ts` to allow the datasources to be passed in
- Updated the `src/index.ts` to instantiate/initialize the data sources and pass them along
- Updated the `src/router.ts` to allow the logger, cache, and datasources through to the controllers
- Consolidated the way we were handling the cache. We were passing the `Cache` itself sometimes and the Cache's `adapter` other times.
- Updated the `controllers` and `services` to use the logger and cache that are now passed in as part of the Request

While trying to manually delete tables, I found that the `users` and `affiliations` tables had circular foreign key relationships.

- Added a new data migration to drop the foreign keys from `users` and `affiliations`

I commented out all of the resolver tests for this PR as I did not want it to get too big! I will be doing the resolvers in a separate PR

- Started refactoring the Model mocks to work with the test DB
- Commented out all resolver tests

## Type of change

## How Has This Been Tested?

Unit tests are passing. Was able to connect and login via the frontend UI. poked around a bit and things seemed to be working alright

## Checklist:

### DIFF
--- a/.env-example
+++ b/.env-example
@@ -58,9 +58,20 @@ MYSQL_CONNECTION_LIMIT=10
 MYSQL_QUEUE_LIMIT=100
 MYSQL_CONNECT_TIMEOUT=60000 # 1 minute
 
+# TEST DATABASE CONFIG
+MYSQL_TEST_HOST=test-mysql        # Defined in docker-compose.yml
+MYSQL_TEST_PORT=3308              # Defined in docker-compose.yml
+MYSQL_TEST_DATABASE=dmsp_test     # Defined in docker-compose.yml
+MYSQL_TEST_USER=root              # Defined in docker-compose.yml
+MYSQL_TEST_PASSWORD=testing123    # Defined in docker-compose.yml
+
 # DynamoDB Table settings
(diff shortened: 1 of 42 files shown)


jupiter007 (approved):
Nice work. LGTM.

PR #329, opened 2025-06-25
Updates to logger

briri:
Small tweaks to the pino logger format to try and get `level`, `jti`, `requestId` and `userId` up to the top level so that we can query and filter them within OpenSearch.

Here is what it looks like curently:
```
{
  "_index": "dmp-log-dev-apollo",
  "_id": "1%3A0%3AnfGYp5cB5-reP3Ko0-Pc",
  "_version": 1,
  "_score": null,
  "_source": {
    "@timestamp": "2025-06-25T14:58:19.281Z",
    "container_name": "apolloServer",
    "source": "stdout",
    "log": "{\"level\":20,\"time\":1750863499281,\"pid\":1,\"hostname\":\"ip-10-66-6-124.us-west-2.compute.internal\",\"requestId\":\"7d88f4a722c710852e3b88e67dbce153\",\"jti\":\"17-1750863491338\",\"userId\":17,\"msg\":\"me resolver, sql: SELECT * FROM users WHERE id = ?, vals: 17\"}",
    "container_id": "609aebcc772448748e657cdfd3af5eee-0435958461",
    "ecs_cluster": "dmp-tool-dev-ecs-cluster-Fargate",
    "ecs_task_arn": "arn:aws:ecs:us-west-2:671846987296:task/dmp-tool-dev-ecs-cluster-Fargate/609aebcc772448748e657cdfd3af5eee",
    "ecs_task_definition": "dmp-tool-dev-ecs-apollo-apollo-server:54"
  },
  "fields": {
    "@timestamp": [
      "2025-06-25T14:58:19.281Z"
    ]
  },
  "sort": [
    1750863499281
  ]
}
```


jupiter007 (approved):
LGTM.

I just had a question about using `return logger.child(metadata)` every time we log stuff. Also, wondering if we should change the name`formatLogMessage` to something else that indicates we are returning a logger instance.

briri:
> LGTM.
> 
> I just had a question about using `return logger.child(metadata)` every time we log stuff. Also, wondering if we should change the name`formatLogMessage` to something else that indicates we are returning a logger instance.

Yeah, that's a good point. I just read through the pino docs again and I am not doing this efficiently. I should probably be creating the child in the `buildContext` function which gets called on an incoming request, and then have the `formatLogMessage` function use the one in the context directly (or there might not be a need for that function at all).

I will do some refactoring

briri:
ok @jupiter007 I refactored the logger code. The `child()` function is only called when necessary now. I ended up replacing the `formatLogMessage` function with a `prepareObjectForLogs` and then went through and updated everywhere the old function was called

Issue #328, opened 2025-06-25
[Bug] - Backend is returning error `Cannot read properties of null (reading 'errors')` when trying to publish a template


jupiter007:
The `Publish template` functionality on the `/template/[templateId]` page recently broke - possibly due to some changes related to versionedQuestions `questionTypeId` field being specified as `NOT NULL`.

<img width="1299" alt="Image" src="https://github.com/user-attachments/assets/8b795307-3297-4e2f-92dd-31a9cc2fce84" />

Steps to reproduce:
1. Log in on `dev` and navigate to https://dmptool-dev.uc3dev.cdlib.net/en-US/template
2. Click `Publish template` button, and check `Visibility setting`, and enter `changelog` and click `Save and publish` button. We always get the `Cannot read properties of null (reading 'errors')` error. 

From quickly looking over backend errors, it appears that the required `questionTypeId` in versionedQuestions is causing the problem when we call `createTemplateVersion`

briri:
this can be tested by successfully publishing a template

PR #334, opened 2025-07-09
Fixed a template publish error

jupiter007:
## Description

Fixed bug where clients calling `createTemplateVersion` in the `template` resolver would get an error when trying to publish because adding data to `versionedQuestions` table required that `questionTypeId` not be `null`. I added a `data-migration script` to allow `null` because we now store the `question type` in the `json` field do not require `questionTypeId` field to tell us which question type the question is using.

Fixes # ([328](https://github.com/CDLUC3/dmsp_backend_prototype/issues/328))

## Type of change

## How Has This Been Tested?
Tested manually to make sure we could publish a template now.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,6 +172,7 @@
 - Old DMPHubAPI datasource and renamed DMPToolAPI to DMPHubAPI since that one had all of the new auth logic
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


sfisher (approved):
This is a good change to allow NULL. Thanks for updating it.

jupiter007:
> This is a good change to allow NULL. Thanks for updating it.

Thanks Scott.

Issue #308, opened 2025-06-13
When a Plan is created, make the owner the initial Primary Contact


bofstein:
They can change it later in the UI, this is just the default.

jupiter007:
@bofstein it looks like in order for the owner to be the initial Primary Contact, they also have to be a Project member already. Currently, we don't add the owner as a `projectMember` when they create a project. Should we?

bofstein:
Yes - we decided that in that last [Backend Bonanza meeting](https://docs.google.com/document/d/1gTgFlOJu9b7EvCLs1HffzNrRoOByUi2NdEEPDxX-II0/edit?tab=t.6p10578rfxxy) that I thought I made a ticket for but may have missed. 

> > - When Project is created, Owner should become a Member and the default Primary Contact
> - For now, add field in backend for Primary Contact at Project level, which should be the Owner, but don’t use it (yet) in the Front end
> - Owners cannot remove themselves as members from a Project, only from Plans
> - When Plan is created, make Owner the initial Primary Contact
> - Allow all Project members to be selected as Primary Contact, not just Plan members

bofstein:
Ah here is that ticket: https://github.com/CDLUC3/dmsp_backend_prototype/issues/307

jupiter007:
Thanks Becky. Maybe I'll work on ticket #307 first, since this one depends on that change.

Issue #307, opened 2025-06-13
When a Project is created, the owner should become a permanent Project Member (contributor) and the Primary Contact


bofstein:
We don't plan to show Primary Contact on the UI but we'll keep it on the backend for now.

The Owner cannot be removed as a Project Member.  They can be removed from specific Plans, but not the whole Project.

PR #339, opened 2025-07-17
Add logic to set the default primary contact

### LINKED ISSUE #244: Allow for default isPrimaryContact to be set
When you are creating a project, we will set a default primary contact.

Whoever is creating it will become the default contact.

Add isPrimaryContact column to projectContributors table.

briri:
## Description

Fixes #244, #307, and #308

**Be sure to run `data-migrations/process.sh` to apply the DB changes!**

I found a bunch of bugs as well while testing that I've addressed and listed below

### Primary Contact
- Added `projectMembers.isPrimaryContact` field to DB and `ProjectMember` model and GraphQL schema
- Added `setCurrentUserAsProjectOwner` and `ensureDefaultProjectContact` functions to the `projectService` and updated the `project` resolver to call them.
- Added the new `ensureDefaultPlanContact` function to the `planService` module and updated the `plan` resolver to use it
- Added a `sendEmailNotification` argument to the `create` function of `ProjectCollaborator` (defaults to true) so that we can bypass sending an email when setting the project creator as the default contact.
- Replace old `ProjectMember.findPrimaryByPlanId` function with `ProjectMember.findPrimaryContact`
- Added `PlanMember.findPrimaryContact` function
- Added `return formatISO9075(new Date(parent.blah));` to all resolvers so that dates are always in the correct format
- Added `invitedBy` to the `ProjectCollaborator` resolver
- Updated the `project` resolver and schema so that `searchExternalProjects` has its own `input` type definition

### Bug Fixes
- Removed self referential foreign key constraints from the `users` table (makes it impossible to delete user records)
- Removed `NOT NULL` constraint from the `users.affiliationId` field (makes it hard to create an initial user record)
- Dropped the `versionedQuestions.questionTypeId` field and foreign key constraints (missed this one when those changes were made)
- SQL script to convert `affiliations.types` to upper case to work with the `AffiliationType` enum
- Exposed the DynamoDB port in the local `docker-compose.yaml` so that AWS CLI commands can be run against it
- Discovered some intermittent issues with the new MySQL connection pool. Changed `keepAliveInitialDelay` to zero and also added an `initPromise` to the class so that we can effectively wait for the connection pool to finish initializing before attaching it to the Apollo server context
- Added a `validateConnection` function to the mysql datasource
- Updated the `index.ts` to use the new `initPromise` and also added some missing `process.exit` statements
- Added time constraints to the `Affiliation.top20` function so that it is inclusive of the current day
- Found that `Collaborator` classes were not awaiting the call to `super` in their `isValid` functions
- Fixed an issue where `isValid` was being called too soon in the `Collaborator` classes
- Fixed an issue with `PlanMember` classes' `findByProjectMemberId` function so that it returns an empty array instead of null
- Fixed some typos in `MemberRole` SQL queries
- Added `sqlStatement` to the log output when `MySQLModel.query` fails
- Fixed a bug where `Plan.isPublished` was not returning the correct result
- Fixed a bug where `Plan.visibility` was not being defaulted to `PRIVATE`
- Fixed a bug on `Project` model that was not allowing SuperAdmins or Admins to fetch a Project
- Fixed a bug where `ProjectSearch` was still trying to fetch the user's email from the old table
- Fixed a bug that was preventing `ResearchDomain.create` from working because `parentResearchDomain` is an unknown field
- Fixed a bug on `VersionedSection` so that `findBySectionId` and `findByTemplateId` return an empty array instead of null when no results are found
- Fixed a bug on the `affiliation` resolver that was incorrectly checking the record's provenance
- Fixed some bugs in the permissions checks for `collaborator`, `member`, `funding`, `project` and `plan`
- Fixed an issue with the `planFunder` schema that was referencing `project` instead of `plan`
- Fixed some auth check bugs in `templateService`

### Revert test MySQL DB and DynamoDB Table
- Removed test MySQL from the docker-compose.yaml
- Removed test DB values from `.env.example`
- Removed logic that used the test MySQL DB in the `data-migrations/nuke-db.sh` and `data-migrations/process.sh` scripts

## Type of change

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/src/datasources/mysql.ts
+++ b/src/datasources/mysql.ts
@@ -62,12 +62,8 @@ export class MySQLConnection implements DatabaseConnection {
 
   // Verify that the pool is able to establish a connection
   private async validateConnection(): Promise<void> {
-    try {
-      const connection = await this.pool.getConnection();
-      connection.release();
-    } catch (err) {
-      throw new DatabaseError('Failed to validate initial connection', err);
-    }
+    const connection = await this.getConnection();
+    connection.release();
   }
 
   // Get a new connection
@@ -108,7 +104,7 @@ export class MySQLConnection implements DatabaseConnection {
       throw new DatabaseError('Database query failed', err);
     } finally {
       if (connection) {
-        await connection.release();
+        connection.release();
       }
     }
   }
--- a/src/services/templateService.ts
+++ b/src/services/templateService.ts
@@ -12,9 +12,10 @@ import { prepareObjectForLogs } from "../logger";
 export const hasPermissionOnTemplate = async (context: MyContext, template: Template): Promise<boolean> => {
   if (!context || !context.token) return false;
 
+  // If the user is a super admin they have access
   if (isSuperAdmin(context.token)) return true;
 
-  // If the current user belongs to the same affiliation OR the user is a super admin
+  // If the current user belongs to the same affiliation
   if (context.token?.affiliationId === template?.ownerId) {
     return true;
   }


src/datasources/mysql.ts
```diff
@@ -35,24 +36,47 @@ export class MySQLConnection implements DatabaseConnection {
...
-        err instanceof Error ? err : undefined
-      );
+      throw new DatabaseError('Failed to create connection pool', err);
     }
   }
 
+  // Verify that the pool is able to establish a connection
+  private async validateConnection(): Promise<void> {
+    try {
+      const connection = await this.pool.getConnection();
```
jupiter007, line 66: L66: Instead of calling `this.pool.getConnection()` would it make sense to use `this.getConnection()`

src/datasources/mysql.ts
```diff
@@ -35,24 +36,47 @@ export class MySQLConnection implements DatabaseConnection {
...
+      connection.release();
+    } catch (err) {
+      throw new DatabaseError('Failed to validate initial connection', err);
```
jupiter007, line 69: L68: Do we need a try-catch here if `getConnection()` is already handling errors?

src/datasources/mysql.ts
```diff
@@ -79,28 +104,23 @@ export class MySQLConnection implements DatabaseConnection {
...
-      throw new DatabaseError(
-        'Database query failed',
-        err instanceof Error ? err : undefined
-      );
+      context.logger.error(prepareObjectForLogs({ sql, values, err }), 'Unable to process SQL query');
+      throw new DatabaseError('Database query failed', err);
     } finally {
       if (connection) {
-        connection.release();
+        await connection.release();
```
jupiter007, line 111: L111: I don't think you need the `await` for a `connection.release()`

src/services/templateService.ts
```diff
@@ -12,8 +12,10 @@ import { prepareObjectForLogs } from "../logger";
 export const hasPermissionOnTemplate = async (context: MyContext, template: Template): Promise<boolean> => {
   if (!context || !context.token) return false;
 
+  if (isSuperAdmin(context.token)) return true;
+
   // If the current user belongs to the same affiliation OR the user is a super admin
-  if (context.token?.affiliationId === template?.ownerId || await isSuperAdmin(context.token)) {
+  if (context.token?.affiliationId === template?.ownerId) {
```
jupiter007, line 18: L17: Minor but you can remove the `OR the user is a super admin` from the comment

jupiter007 (approved):
Hi Brian. Looks good to me. I only had a few comments.

src/datasources/mysql.ts
briri: No. I probably had it as a public function at one point. Will remove it as it's only called internally now

Issue #322, opened 2025-06-18
Inconsistencies between the Zod schema from `dmptool-types` and the json schemas for those question types in the db


jupiter007:
While working on the question types frontend ticket ([#157](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/157)), I noticed that the Type definitions supplied by `dmptool-types` for `numberRange` and `dateRange` do not match the `json schema` coming from the `questionTypes` table. Namely it looks like `min`, `max` and `step` are missing from the type definitions generated by the Zod schema.

Also, I noticed that the `CurrencyQuestionType` from `dmptool-types` places `denomination` under `meta`, but the `json` for `currency` `questionTypes` places it under `attributes`. 

Additionally, it might be good to add question `label` and `help` properties for all question types so that users can save that info to accompany the question. For example, the `Affiliation Search` question type provides a way to save this info, but most other question types do not.

bofstein:
Is there a way for me to test this?

jupiter007:
> Is there a way for me to test this?

I think if you just confirm that question types `dateRange`, `numberRange` and `currency` work, that is sufficient.

PR #354, opened 2025-07-23
Update code to work with new @dmptool/types v1.2.0

briri:
## Description

**These changes will break the frontend UI!!!** A PR for the frontend will be coming soon.

Associated with changes made for #322
- Updated `@dmptools/types` dependency
- Drop the QuestionType schema, resolver and Model
- Added data migration to drop the old `questionTypes` table since the info is now in the above package
- Added data migration to update (or delete in a few of the more complex cases) question test records

## Type of change

## How Has This Been Tested?

All tests are passing and validated through the frontend UI

## Checklist:


src/resolvers/question.ts
```diff
@@ -82,6 +82,8 @@ export const resolvers: Resolvers = {
           // create the new question
           const newQuestion = await question.create(context);
 
+console.log(newQuestion);
```
jupiter007, line 85: Did you mean to keep this console.log()?

jupiter007 (approved):
LGTM

Issue #373, opened 2025-08-05
Backend is returning the wrong "totalCount" when querying "myProjects" with "paginationOptions"


jupiter007:
Currently, the "myProjects" `paginatedQueryByCursor` is returning a `totalCount` of `0`.

briri:
I think I have a fix for this. Discovered it while working on #368

PR #374, opened 2025-08-05
[Bug fix] - totalCount returned as 0, or was changing with each request…

jupiter007:
## Description

While working on the frontend ticket [647](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/647), I noticed that when making requests for myProjects, I was getting a totalCount =0 even though there was more than one item found from a search.

- I updated `paginatedQueryByCursor` to use the original `whereClause` when calling `getTotalCountForPagination` because otherwise,  the `totalCount` was changing on each request.
- Made update to return `countResponse` differently based on whether there was a `groupBy` clause. Previously, the totalCount returned was incorrect, and looked something like the following, with each object representing one found item in the query due to the `groupBy` clause
```
[{ total: 3 },{ total: 3 },{ total: 1 }]
```

Fixes # ([373](https://github.com/CDLUC3/dmsp_backend_prototype/issues/373))

## Type of change

## How Has This Been Tested?
Manual test and unit test

## Checklist:
* Note: Not a dependency but I am working on a frontend ticket that needs the correct totalCount (#[647](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/647))

### DIFF
--- a/src/models/MySqlModel.ts
+++ b/src/models/MySqlModel.ts
@@ -182,8 +182,16 @@ export class MySqlModel {
       const fromClause = sqlParts[sqlParts.length - 1];
 
       const countSql = `SELECT COUNT(${countField}) total FROM ${fromClause} ${whereClause} ${groupByClause}`;
+
       const countResponse = await MySqlModel.query(apolloContext, countSql, values, reference);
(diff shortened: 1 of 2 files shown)


briri (approved):
Thanks @jupiter007 yes, this looks like it will work

Issue #366, opened 2025-08-04
Update backend so that it returns whether the question was answered or not for the publishedQuestions query


jupiter007:
Currently, in order for the frontend to figure out whether a question on the `Plan Section` page was answered, it would have to do an additional query for each question using the `versionedSectionId` and `versionedQuestionId` returned from the `publishedQuestions` query.

It would make a lot more sense for the backend to just return that info in the one `publishedQuestions` query.

We can update the backend query to accept a `planId`

<img width="1908" height="1280" alt="Image" src="https://github.com/user-attachments/assets/0e030c49-80be-4d57-a331-ad0f8aaf54c4" />

bofstein:
Confirmed I'm now seeing correct Questions Answered count on Section and Project overview

PR #380, opened 2025-08-11
update backend so that it returns whether the question was answered or not for the publishedquestions query

sfisher:
## Description

https://github.com/CDLUC3/dmsp_backend_prototype/issues/366

- Adds a boolean of `hasAnswer` to the resolver data so that it indicates whether the question is answered.
- Adds a finder to the Answer to find the filled answers for a list of versionQuestions.
- Logic is currently simple (there is a record, not null, not blank) for the question since we do not yet know how specific we need to get or if we need to check different JSON values for a whole different list of question types.  I suspect it may not be super useful or easy to know answered for all types, anyway.  Like if there are lists of checkboxes and they have not checked any, they may have done so deliberately since it's often a valid state to not check any checkboxes (not radio buttons).
- Non-breaking: added new `publishedQuestionsWithAnsweredFlag` but kept existing `publishedQuestions`.

I added tests for the changes to `answers` but not clear how to test this resolver right now since tests don't really exist for resolvers right now.  I think this was something Brian was working through.

Fixes # (366)

## Type of change

## How Has This Been Tested?

- Added test for findAnswersByQuestionIds finder in Answer.
- Verified that the resolver returns appropriate data from GraphQL and marks items correctly with `hasAnswer` boolean.
Screenshot 2025-08-08 at 4 29 26 PM
- Testing for the resolvers is unclear right now and the test file for this resolver doesn't currently exist.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 ## v0.2 - Initial deploy to the stage environment
 
 ### Added
+- Added `findFilledAnswersByQuestionIds` which takes versionedQuestionIds and returns the answers for those questions.
 - Added data migration to clean up old question JSON so it conforms with new @dmptool/types schemas
 - Added data migration to drop the old `questionTypes` table
 - Added `updatePlanFunding` to allow the update of multiple `planFunding` records [#305]
--- a/src/models/Answer.ts
+++ b/src/models/Answer.ts
@@ -142,7 +142,7 @@ export class Answer extends MySqlModel {
   }
 
   // given a list of question ids, return all filled answers, flexible to handle different groupings of question ids
-  static async findFilledAnswersByQuestionIds(
+  static async findAnswersByQuestionIds(
     context: MyContext,
     questionIds: number[]
   ): Promise<Answer[]> {
@@ -150,7 +150,7 @@ export class Answer extends MySqlModel {
 
     const placeholders = questionIds.map(() => '?').join(', ');
     const sql = `SELECT * FROM answers WHERE versionedQuestionId IN (${placeholders}) AND json IS NOT NULL AND json != ''`;
-    const results = await Answer.query(context, sql, questionIds.map(String), 'Answer.findFilledAnswersByQuestionIds');
+    const results = await Answer.query(context, sql, questionIds.map(String), 'Answer.findAnswersByQuestionIds');
     return Array.isArray(results) && results.length > 0 ? results.map((ans) => new Answer(ans)) : [];
   }
 };
--- a/src/models/Answer.ts
+++ b/src/models/Answer.ts
@@ -148,7 +148,8 @@ export class Answer extends MySqlModel {
   ): Promise<Answer[]> {
     if (!Array.isArray(questionIds) || questionIds.length === 0) return [];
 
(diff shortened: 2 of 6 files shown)


jupiter007 (approved):
@sfisher it looks fine, but I'm wondering if it's ok to use the existing `publishedQuestions` query and update it to pass the `planId` variable and return `hasAnswer`. I believe in the couple of scenarios where the frontend calls `publishedQuestions` we will need `hasAnswer` anyways.

I'm not too concerned about backwards compatibility, since it's only effecting a couple of pages on the frontend and we'll have to update the frontend in any case to get the new `hasAnswer` prop . Also, the apps aren't live yet, so it's less of a concern.

Lastly, it seems like this is adding a lot of duplicated code if we use both `publishedQuestions` and `publishedQuestionsWithAnsweredFlag` and also with type `VersionedQuestion` and `VersionedQuestionWithFilled`.

Not critical so I'm approving, but just expressing my preference.

sfisher:
Ok.  I can go bak and update this.

sfisher:
@jupiter007 .  There aren't places where you want to get the questions without knowing if they're filled?  Like when someone's designing questions? Or printing out just the templates without referring to a plan?  Or maybe those aren't "published" questions?

I mean I guess I could also make plan optional and then if it's not supplied just set everything to false if there is no plan id to use in looking anything up.

sfisher:
Thank you, Juliet.  This all makes sense.

I've updated to only have the resolver for `VersionedQuestions.publishedQuestions` and it always needs a `planId` and also returns the data structure with the additional hasAnswer boolean.

The thing I was confused about were the return types in the schema for `VersionedQuestion` since I assume someone will eventually need to look at `VersionedQuestion` that is not associated with a plan since I see that they need to be associated with versionedTemplates, but someone creating a template doesn't need to have a plan yet (as far as I can tell).

So I think there are reasons to keep both data structures (VersionedQuestion and VersionedQuestionWithFilled) and will be used in different circumstances as the return values.

I find the conflation of GraphQL types and Typescript types fairly confusing since they are similar but different and the generation automatically creates typescript types for GraphQL types.  And internal typescript types that aren't GraphQL types get defined somewhere else or is everything both a GraphQL type and a typescript type?  😕

jupiter007:
> Thank you, Juliet. This all makes sense.
> 
> I've updated to only have the resolver for `VersionedQuestions.publishedQuestions` and it always needs a `planId` and also returns the data structure with the additional hasAnswer boolean.
> 
> The thing I was confused about were the return types in the schema for `VersionedQuestion` since I assume someone will eventually need to look at `VersionedQuestion` that is not associated with a plan since I see that they need to be associated with versionedTemplates, but someone creating a template doesn't need to have a plan yet (as far as I can tell).
> 
> So I think there are reasons to keep both data structures (VersionedQuestion and VersionedQuestionWithFilled) and will be used in different circumstances as the return values.
> 
> I find the conflation of GraphQL types and Typescript types fairly confusing since they are similar but different and the generation automatically creates typescript types for GraphQL types. And internal typescript types that aren't GraphQL types get defined somewhere else or is everything both a GraphQL type and a typescript type even if GraphQL doesn't expose it? 😕
> 
> Oh, it's working in Apollo explorer.
> 
> Screenshot 2025-08-13 at 4 56 59 PM

Cool.Thanks for making the change @sfisher.

Issue #243, opened 2025-03-25
Update resolvers for commenting


bofstein:
- Allow Edit and Remove comment functionality (only for your own)
- Get all comments
- Email collaborators for new comments (only Add, not edit or remove)

jupiter007:
Putting into PR review

PR #385, opened 2025-08-19
Added resolvers for comments

jupiter007:
## Description

The backend needs to return both the `feedbackComments` and `answerComments` to the frontend so that the list of comments can be displayed on the Question Details page. The comments are associated with the `answerId`, and the users with appropriate permissions should be able to add, edit and delete comments.

- Added `PlanFeedback` and `PlanFeedbackComment` models 
- Added answer and feedback resolvers
- Added `feedback` chained resolver to `plan`
- Updated associated schemas with new resolvers
- Added `feedbackComments` and `User` chained resolvers to `answer`
- Updated `emailService` to add `projectCollaboratorCommentsAdded` so that we can send emails to project collaborators when a new comment has been added.

Fixes # ([243](https://github.com/CDLUC3/dmsp_backend_prototype/issues/243))

## Type of change

## How Has This Been Tested?
Manually with frontend changes, and with unit tests.

## Checklist:


src/resolvers/answer.ts
```diff
@@ -144,6 +152,115 @@ export const resolvers: Resolvers = {
...
+            throw NotFoundError(`Answer with ID ${answerId} not found`);
+          }
+          const plan = await Plan.findById(reference, context, answer.planId);
+          if (!plan) {
+            throw NotFoundError(`Plan ${answer.planId} not found`);
+          }
+          const project = await Project.findById(reference, context, plan.projectId);
+          if (await hasPermissionOnProject(context, project)) {
+            // Send out email to project collaborators to let them know that comments were added
+            await sendProjectCollaboratorsCommentsAddedEmail(context, context.token.email, context.token.id)
```
briri, line 175: you're sending the notification to the user who added the comment. I think we want to send it to the other collaborators.

src/resolvers/answer.ts
```diff
@@ -144,6 +152,115 @@ export const resolvers: Resolvers = {
...
+          const project = await Project.findById(reference, context, plan.projectId);
+
+          if (await hasPermissionOnProject(context, project)) {
+            const answerComment = await AnswerComment.findById(reference, context, answerCommentId);
+
+            if (!answerComment) {
+              throw NotFoundError(`Answer comment ${answerCommentId} not found`);
+            }
+
+            return await answerComment.delete(context);
```
briri, line 256: we probably only want the user who created the comment to delete it. so use `if (answerComment.createdById === context.token.id) {` here as well

src/resolvers/feedback.ts
```diff
@@ -0,0 +1,319 @@
...
+
+          // Check that user has permissions to access feedback
+          const projectId = plan.projectId;
+          const project = await Project.findById(reference, context, projectId);
+          if (!project) {
+            throw NotFoundError(`Project with ID ${projectId} not found`);
+          }
+          if (await hasPermissionOnProject(context, project)) {
+            const temp = await PlanFeedback.findByPlanId(reference, context, planId);
+            console.log("****PLAN FEEDBACK", temp);
```
briri, line 49: remove this debug statement

src/resolvers/feedback.ts
```diff
@@ -0,0 +1,319 @@
...
+          const plan = await Plan.findById(reference, context, planId);
+          if (!plan) {
+            throw NotFoundError(`Plan with ID ${planId} not found`);
+          }
+          const project = await Project.findById(reference, context, plan.projectId);
+          if (!project) {
+            throw NotFoundError(`Project with ID ${plan.projectId} not found`);
+          }
+
+          if (await hasPermissionOnProject(context, project)) {
```
briri, line 109: maybe a question for @bofstein.

The current tool only allows the owner of the plan (or a co-owner) to request feedback. We are allowing anyone with edit access to do so here. Do we want that or should we restrict it to owner(s)?

src/models/PlanFeedback.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  // Fetch a PlanFeedback by it's id
+  static async findById(reference: string, context: MyContext, feedbackId: number): Promise<PlanFeedback> {
+    const sql = `SELECT * FROM ${PlanFeedback.tableName} WHERE id = ?`;
+    const results = await PlanFeedback.query(context, sql, [feedbackId?.toString()], reference);
+    return Array.isArray(results) && results.length > 0 ? new PlanFeedback(results[0]) : null;
+  }
+
+  // Fetch PlanFeedback by it's planId and requestedById
+  static async findByPlanIdAndRequestedById(
```
briri, line 104: If the plan allows for multiple rounds of feedback, then a single user can potentially request feedback more than once, so this should return an array instead of a single record.

src/models/PlanFeedback.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+    // Otherwise return as-is with all the errors
+    return new PlanFeedback(this);
+  }
+
+  //Update an existing PlanFeedback
+  async update(context: MyContext, noTouch = false): Promise<PlanFeedback> {
+    if (await this.isValid()) {
+      if (this.id) {
+        await PlanFeedback.update(context, PlanFeedback.tableName, this, 'PlanFeedback.update', ['planFeedbackComments'], noTouch);
```
briri, line 67: call `prepForSave` before calling `update`

src/models/PlanFeedbackComment.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+    // Otherwise return as-is with all the errors
+    return new PlanFeedbackComment(this);
+  }
+
+  //Update an existing PlanFeedbackComment
+  async update(context: MyContext, noTouch = false): Promise<PlanFeedbackComment> {
+    if (await this.isValid()) {
+      if (this.id) {
+        await PlanFeedbackComment.update(context, PlanFeedbackComment.tableName, this, 'PlanFeedbackComment.update', [], noTouch);
```
briri, line 55: call `prepForSave` before `update`

briri (changes requested):

src/services/emailService.ts
```diff
@@ -172,3 +176,32 @@ export const sendProjectCollaborationEmail = async (
...
+  email: string,
+  userId?: number,
+): Promise<boolean> => {
+  let toAddress = email;
+  const message = emailMessages.projectCollaboratorCommentsAdded;
+
+  if (userId) {
+    const user = await User.findById('sendProjectCollaborationEmail', context, userId);
+    // Bail out if the user has asked us not to send these notifications
+    if (!user.notify_on_template_shared) {
```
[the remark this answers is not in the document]
jupiter007: Thanks Brian. I realized that I don't need the `userId` in this case, and that I can pass in just the projectCollaborators emails, minus the userId to this function.

src/resolvers/answer.ts
jupiter007: Thanks. I updated the resolvers to send a list of the projectCollaborator emails and filtered out the user's own email.

src/services/emailService.ts
```diff
@@ -172,3 +176,32 @@ export const sendProjectCollaborationEmail = async (
...
 }
+
+export const sendProjectCollaboratorsCommentsAddedEmail = async (
+  context: MyContext,
+  email: string,
+  userId?: number,
+): Promise<boolean> => {
+  let toAddress = email;
+  const message = emailMessages.projectCollaboratorCommentsAdded;
+
```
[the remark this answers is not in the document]
jupiter007: Ok, I updated the function to send to all projectCollaborators

src/models/PlanFeedback.ts
jupiter007: Done

jupiter007:
@briri thanks so much for your review. 

I responded to your comments.

jupiter007:
> > @briri I did have one more question. If there are multiple rounds of feedback going on for the same plan, then when an Admin adds a comment, which feedback record do we associate the comment with? For now, I was just using the first feedback id that came up, but wasn't sure how to handle that.
> 
> There should only ever be one "Active" round of feedback for a plan at any given time. Speaking of that, can you update the resolver so that feedback cannot be requested if there is an active round (one that is not "complete")?
> 
> Then I think you can always just add comments to the active round

Thanks for clarifying Brian. I've added a check for any active `feedback` rounds in the `requestFeedback` mutation.

briri (approved):
Looks good @jupiter007 

I had just one more change on the `Feedback` model to remove the `feedbackComments` as it's not used anywhere. Once that's done feel free to merge

src/models/PlanFeedback.ts
```diff
@@ -0,0 +1,122 @@
...
+import { PlanFeedbackComment } from "./PlanFeedbackComment";
+
+export class PlanFeedback extends MySqlModel {
+  public planId: number;
+  public requested: string;
+  public requestedById: number;
+  public completed?: string; //Optional - won't exist initially when request is made
+  public completedById?: number; //Optional - won't exist initially when request is made
+  public summaryText?: string; //Optional - won't exist initially when request is made
+  public planFeedbackComments?: PlanFeedbackComment[];
```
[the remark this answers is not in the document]
jupiter007: Thanks again for your review. I removed the planFeedbackComments.

PR #400, opened 2025-08-27
Fixing user creation for signup

sfisher:
## Description

See https://github.com/CDLUC3/dmsp_frontend_prototype/issues/708 .  User signup fails with unknown error.

Fixes https://github.com/CDLUC3/dmsp_frontend_prototype/issues/708

This was an odd problem since it seemed to be re-assigning context to a different value with a definition like `const context = ...`.  A weird part about it was that the git history showed this line was created in February and hadn't been changed since then. I'm not sure why it began failing now--maybe because it used to be reassigned after it was used the last time or 🤷   Or maybe it's been failing for a long while and no one noticed until now.

I'm also not sure what the purpose of the line was but removing it seems to fix the problem and I don't see any reason it needed to be there.

Changes to the user.ts file are to make it pass linting even though I didn't touch it in my changes.  Unused imports were present.

## Type of change

## How Has This Been Tested?

Tested signup with new user (with all required fields) and it succeeds in creating user and logging in.

Checked database and expected information is present.

## Checklist:
- [x] ~I have added tests that prove my fix is effective or that my feature works~ I believe there are outstanding tests for these controllers and similar that Brian began working on but seemed to fail and needed additional work to be reliable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - changed `sections` resolver to `versionedSections` on the `src/resolvers/plan.ts` file and changed the reference for `PlanSearchResult.sections` to `versionedSections`
 
 ### Fixed
+- Fixed an issue where signup failed because context had been reset to different object
 - Fixed an issue causing the DMP version to not include sections/questions in the narrative if they had not been answered
 - Fixed an issue with the null/undefined check on model queries that use 'searchTerm'
 - When generating a new `versionedTemplate`, we need to deactivate the old ones in the db [#363]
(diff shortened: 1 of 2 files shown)


briri:
interesting. good catch @sfisher. I didn't fully understand how the Apollo server context worked initially, and so the non-graphql endpoints like signup didn't have one. I did some fairly recent refactoring and updated the way the context was managed outside graphql. That's likely when the issue started.

Anyway, you're right, its getting passed into the function so there's no need for that line

Issue #358, opened 2025-07-29
Backend should always ensure that the project and plan have a member


briri:
The new functions that ensure that the project/plan have at least one member (and primary contact) should be called from the other resolver mutations when the action would result in the project/plan having no members.

For project:
- If the `remove` mutation is called, the owner of the project should become a member

For plan:
- If the `remove` mutation is called and the member being removed was the "primary contact", then the owner (if they are a member) should become the primary contact otherwise assign it to the PI that was created/added first

bofstein:
Confirmed

<img width="1141" height="472" alt="Image" src="https://github.com/user-attachments/assets/3986ba68-3522-41e4-aa2d-5aa9ea4340d1" />

PR #401, opened 2025-08-28
Should always ensure that Project and Plan have a member

jupiter007:
## Description

- Updated `ProjectMember.delete` to check whether there is more than one projectMember, otherwise, we return an error.
-- Also, checked that for all the plans associated with that `projectMemberId` in the `planMembers` table, none of them would be left with no member. If so, we stop the deletion of that member and return an error.
-- We delete the `planMembers` records only if there is more than one planMember left for the associated planIds.
- Updated 'PlanMember.delete` to only delete the planMember if there is more than one planMember left under the given `planId`. If not, then return an error.
- Updated unit tests
- Updated `addProjectMember` resolver to default the new member to a role if none was provided

Fixes # ([358](https://github.com/CDLUC3/dmsp_backend_prototype/issues/358))

## Type of change

## How Has This Been Tested?
Manually tested and with unit tests

## Checklist:
*Note: There is a related frontend ticket to display a modal warning a user when they want to delete a ProjectMember: https://github.com/CDLUC3/dmsp_frontend_prototype/issues/737

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # DMP Tool Apollo Server Change Log
 
 ### Added
+- Added checks in `ProjectMember` and `PlanMember` to prevent the deletion of a member if they are the last one [#358]
+- Added a fallback to set a `default` role if none was provided while adding a new `ProjectMember` [#358]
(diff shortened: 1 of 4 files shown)


briri (approved):
LGTM

PR #403, opened 2025-08-28
Remove expiresIn from contents of JWT payloads

briri:
Not sure if this will actually help or not. I can see that the dmpIds content is there when we sign the JWT. Its getting lost though either in the signing process or in the decoding process in the client.

Did some investigation into why the `dmpIds` are not appearing in the decoded JWT. The suggestion was to try removing the `expiresIn` from the payload because this can cause some parsers to do weird things. We are already including the expiresIn when we call `sign` which should be enough.

### DIFF
--- a/src/services/tokenService.ts
+++ b/src/services/tokenService.ts
@@ -33,13 +33,11 @@ export interface JWTAccessToken extends JwtPayload {
   languageId: string,
   dmpIds: JWTAccessTokenDMPId[],
   jti: string,
-  expiresIn: number,
  }
 
 export interface JWTRefreshToken extends JwtPayload {
   jti: string,
   id: number,
-  expiresIn: number,
 }
 
 // Hash a token before placing it in the cache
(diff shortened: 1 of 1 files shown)


jupiter007:
I can actually see the DMPIds both on the frontend and backend.

But I had to update the database so that there is a projectCollaborator with my `email` address.  Then I could see the dmpIds in the tokenService: ```***DMPIDs [
apollo-server  |   { dmpId: 'https://doi.org/10.11111/2A3B4C', accessLevel: 'COMMENT' }
apollo-server  | ]```
and I checked the payload on the frontend after I got the token from logging in, and I see the same `dmpIds` there:
```***PAYLOAD {
  id: 1,
  email: 'super@example.com',
  givenName: 'Super',
  surName: 'Admin',
  affiliationId: 'https://ror.org/03yrm5c26',
  role: 'SUPERADMIN',
  languageId: 'en-US',
  dmpIds: [
  { dmpId: 'https://doi.org/10.11111/2A3B4C', accessLevel: 'COMMENT' }
],
  jti: '1-1756403536177',
  iat: 1756403536,
  exp: 1758203536
}```

jupiter007 (approved):
It worked for me. I see the `dmpIds` in the payload

briri:
ok. Thanks @sfisher it might be that the user you're logging in with doesn't have any plans. Juliet was able to verify that she's seeing them. I am going to merge this since it seems like we don't need this expiresIn within the JWT payload

sfisher:
Ok.  I was able to get them by modifying my database.  ```"dmpIds":[{"dmpId":"https://doi.org/10.11111/2A3B4C"}]```

Issue #405, opened 2025-08-29
[Bug] - Template visibility is not updated when a user updates visibility while publishing


jupiter007:
Currently, when a new template is created, the `visibility` is set to `ORGANIZATION` in the `template` table by default.

When a user publishes that template, the visibility setting is changed in the `versionedTemplate` record, but the `template` record is not updated with that same visibility.

This posed a problem on the `/template` page in the NextJS app, where we are running a `myTemplates` query to get both `published` and `unpublished` templates, and the `visibility` for all the listed templates are all set to the initial `ORGANIZATION` value.

I believe we should be update the `visibility` value in the corresponding `template` table when a template is published.

Another option would be to add a new resolver that returns both `published` and `unpublished` templates by joining the `templates` and `versionedTemplates` tables.

jupiter007:
I put this into PR but in `DRAFT` mode. Need to check with team to make sure this is correct fix.

jupiter007:
Per my discussions with the team, it was decided that we don't need to keep the `visibility` field for the `templates` table. However, we do want a `latestPublishVisibility` field so that we can easily display this info on the template cards on `/template` page.

- Added a data migration script to update `visibility` to `latestPublishVisibility` in `templates` table.
- Added code to update `latestPublishVisibility` whenever a `template` is published

Submitted for PR review

PR #406, opened 2025-08-29
[Bug fix] - template visibility is not updated

jupiter007:
## Description
When a user publishes that template, the visibility setting is changed in the versionedTemplate record, but the template record is not updated with that same visibility.

This posed a problem on the /template page in the NextJS app, where we are running a myTemplates query to get both published and unpublished templates, and the visibility for all the listed templates are all set to the initial ORGANIZATION value.

I believe we should be update the visibility value in the corresponding template table when a template is published.

- Updated createTemplateVersion to update visibility in the corresponding `templates` table record.

Fixes # ([405](https://github.com/CDLUC3/dmsp_backend_prototype/issues/405))

## Type of change

## How Has This Been Tested?
Manually tested with frontend

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - changed `sections` resolver to `versionedSections` on the `src/resolvers/plan.ts` file and changed the reference for `PlanSearchResult.sections` to `versionedSections`
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


briri (approved):
This all looks good @jupiter007. Thanks for making the changes

PR #412, opened 2025-09-03
Updates to README

briri:
Found a couple of issues with the README.
- Updated description of the DynamoDB partition key and sort key structure along with info about how to whether a DMP Id (DOI) has been registered/minted/published with EZID/DataCite.
- Updated the instructions for using AWS CLI to query the local DynamoDB table


jupiter007 (approved):
LGTM

Issue #404, opened 2025-08-29
Answering a Plan question does not trigger new version of DMP


briri:
The `answer` resolver is not generating a new plan version. It should call `await addVersion(context, plan, reference);` after an answer has been successfully added or updated. 

See the `member` or `funding` resolver for examples.

bofstein:
I know this bug was underlying a lot of issues but is there any specific test I can do to confirm?

bofstein:
The CSV bug specifically is fixed now that was supposed to be due to this so I'm fine closing it

PR #413, opened 2025-09-04
Fix issue with commonStandardService management of the latest version of a DMP

briri:
## Description

Part one of fix for #404 

The `commonStandardService` was not properly managing the latest version of the DMP. When the `updateVersion` on the `PlanVersion` model was called, it was correctly generating a snapshot (point-in-time) copy of the current DMP. It was not however updating the latest version.

## Type of change

## How Has This Been Tested?

All unit tests are passing. Validated the changed JSON is correct via the new narrative generator service

## Checklist:

### DIFF
--- a/src/models/PlanVersion.ts
+++ b/src/models/PlanVersion.ts
@@ -75,6 +75,7 @@ export const updateVersion = async (
 
   // If the lastModified date is not within the last hour, create a new version snapshot
   const mostRecentVersion = await latestVersion(context, plan, reference);
+
   if (mostRecentVersion) {
     const lastModified = new Date(mostRecentVersion?.modified);
     const now = new Date();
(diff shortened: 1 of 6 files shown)


src/models/PlanVersion.ts
```diff
@@ -29,6 +30,33 @@ import { Plan } from "./Plan";
...
+
+  const errMsg = "Unable to create a new version snapshot";
+  try {
+    const snapshot = await createDMP(context, dmpId, latestVersion, modified);
+    if (isNullOrUndefined(snapshot)) {
+      context.logger.error(prepareObjectForLogs(latestVersion), `${reference} - Dynamo table createDMP failed`);
+      return errMsg;
+    }
+  } catch (err) {
+    context.logger.error(prepareObjectForLogs({ err, latestVersion }), `${reference} - Fatal error for Dynamo createDMP`);
```
jupiter007, line 54: Are returned errors always going to be in key/value pairs for `prepareObjectForLogs`? Also, `createDMP` logs an error already when it fails, and then we throw again and log again. Would this be redundant?

jupiter007 (approved):
LGTM

PR #431, opened 2025-09-09
Pin some dependencies and upgrade the rest

briri:
Pinning some dependencies due to [this issue](https://www.aikido.dev/blog/npm-debug-and-chalk-packages-compromised).

Upgrading all the rest.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,20 +77,6 @@
         "npm": ">=10.2.4"
       }
     },
-    "node_modules/@ampproject/remapping": {
-      "version": "2.3.0",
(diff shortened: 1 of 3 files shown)


jupiter007:
@briri it looks like the bad package versions were removed, so I think if we just do a rebuild in the pipeline to get the safe versions for now, it's good.

Going forward, we might want to lock down any patches because of possible future attacks?

jupiter007:
Actually, it's a good idea to delete the package-lock.json file and re-run the install. I'll do the same for frontend.

PR #439, opened 2025-09-16
Chore/redact logs

briri:
## Description

Final part of [#14](https://github.com/CDLUC3/dmptool-infrastructure/issues/14)

- Add `redact` functionality to Pino logger
- Add logic to mask sensitive information from the `prepareObjectForLogs` function
- Made sure all logging masks emails like `t****r@example.com`

## Type of change

## How Has This Been Tested?

All tests pass and did some basic testing through the UI locally

## Checklist:

### DIFF
--- a/src/models/MySqlModel.ts
+++ b/src/models/MySqlModel.ts
@@ -218,7 +218,15 @@ export class MySqlModel {
       const vals = values.map((entry) => this.prepareValue(entry, typeof (entry)));
 
       try {
-        apolloContext.logger.debug(prepareObjectForLogs({ sql, values }), reference);
+        // Mask all of the values like 'a***e' before logging
+        const maskedValues = vals.map((val) => {
+          if (typeof val === 'string' && val.length > 3) {
+            const mask = val.slice(1, val.length - 2).replace(/./g, '*');
+            return `${val[0]}${mask.slice(0, 10)}${val[val.length - 1]}`;
+          }
+          return val;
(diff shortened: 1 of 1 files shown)


src/models/MySqlModel.ts
```diff
@@ -215,11 +215,10 @@ export class MySqlModel {
     // The dataSource, logger and sqlStatement are required so bail if they are not provided
     if (dataSources && logger && dataSources.sqlDataSource && sqlStatement) {
       const sql = sqlStatement.split(/[\s\t\n]+/).join(' ');
-      const logMessage = `${reference}, sql: ${sql}, vals: ${values}`;
       const vals = values.map((entry) => this.prepareValue(entry, typeof (entry)));
 
       try {
-        apolloContext.logger.debug(logMessage);
+        apolloContext.logger.debug(prepareObjectForLogs({ sql, values }), reference);
```
jupiter007, line 221: One thing I'm concerned about here is could there be `values` passed that are also `PII` in the queries. If so, then the current redactSensitiveInfo will not capture these.

briri:
> One thing I'm concerned about here is could there be values passed that are also PII in the queries. If so, then the current redactSensitiveInfo will not capture these.

Yeah, good point. I updated the MySQL function to mask all of the values (if they're strings ... can't think of a scenario where any other types could be PII for our use cases). We can update later if we want to be more nuanced than that

PR #450, opened 2025-09-25
Fixed breaking build

jupiter007:
## Description

-Updated Dockerfile.aws to remove the "NODE_ENV=production" from the `Dependencies` stage, since we need devDependencies for the `Build` stage, and we were getting the error:
```

> [builder 5/7] RUN npm run build:
--
0.403
0.403 > dmsp_apollo@1.0.0 build
0.403 > rm -rf ./dist && npm run generate && npm run compile
0.403
0.505 npm warn config production Use `--omit=dev` instead.
0.540
0.540 > dmsp_apollo@1.0.0 generate
0.540 > graphql-codegen
0.540
0.543 sh: graphql-codegen: not found


```

## Type of change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 - changed `sections` resolver to `versionedSections` on the `src/resolvers/plan.ts` file and changed the reference for `PlanSearchResult.sections` to `versionedSections`
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


briri (approved):
That makes sense. Thanks for debugging and fixing @jupiter007

PR #451, opened 2025-09-25
Support for searching for potential collaborators

### LINKED ISSUE #252: Add backend updates to Search for Project Members
Wireframe for Project Member search: 
https://www.figma.com/proto/RRtoRLyr5Spf0KiyRfEclt/DMP-Tool-Wireframes?node-id=183-522&t=M9JzcqhvxBJdTW3Q-0&scaling=min-zoom&content-scaling=fixed&page-id=177%3A504&starting-point-node-id=177%3A642

We need to add a search resolver that will search existing collaborators, other people in the user’s org, and then ORCID from our db, and then later query ORCID API. We will display results in that order as well (e.g., showing an existing collaborator at the top and then any users at their org that match, and finally ORCiD matches).

There was a suggestion that we will need more data about the users, like what OPIDoR does. Fraser may need to make some updates to this wireframe.

The corresponding frontend ticket for this is: #251 .

briri:
## Description

Fixes #252

@jupiter007, @jdddog and @sfisher  **Please Note that you must add ORCID_CLIENT_ID and ORCID_CLIENT_SECRET to your local `.env` file**

- Add new ENV variables `ORCID_CLIENT_ID`, `ORCID_CLIENT_SECRET`, `ORCID_BASE_URL`, `ORCID_AUTH_PATH` and `ORCID_READ_ONLY_SCOPE`.
- Added `src/config/ocidConfig` to provide access to the above ENV variables and also defaults for some of them
- Added new `src/datasources/orcid` to support interactions with the ORCID API
- Added new functions to the `ProjectCollaborator` model:
  - The `findPotentialCollaboratorByORCID` provides support for finding a person by their ORCID. It queries the `User` table first and if nothing is found it makes a call out to the ORCID API.
  - The `findPotentialCollaboratorsByTerm` provides support for find a person by a search term matched against their `email`, `givenName` or `surName`. It looks only for users associated with the current user's organization or those that have collaborated on the user's past projects.
- Added a `findByOrcid` function to the `User` model
- Updated the `findCollaborator` resolver to allow for searching by ORCID or by a search term. If the `term` specified is a valid ORCID, it will fetch the matching user from either the users table or the ORCID API. If a term (greater than 3 characters) is specified it will try to match a user's `email`, `givenName` or `surName` to the specified term and return paginated results (only users associated with the current user's affiliation or who were collaborators on a prior project). If the user is a Super admin, it will simply search across all users

## Type of change

## How Has This Been Tested?

Added new Unit tests and all tests are passing. Tested resolver in Apollo explorer

## Checklist:


src/models/Collaborator.ts
```diff
@@ -376,4 +397,159 @@ export class ProjectCollaborator extends Collaborator {
...
+          orcid: user.orcid || '',
+          email: await user.getEmail(context),
+          affiliationName: affiliation?.name,
+          affiliationRORId: affiliation?.uri,
+          affiliationURL: affiliation?.homepage,
+        };
+
+      } else {
+        // Finally, call the ORCID API to get the person's details
+        const orcidAPI: OrcidAPI = await new OrcidAPI({cache: context.cache});
```
jupiter007, line 436: L436: I don't think you need to `await` setting the new instance of the OrcidAPI here `await new OrcidAPI`.

src/models/Collaborator.ts
```diff
@@ -376,4 +397,159 @@ export class ProjectCollaborator extends Collaborator {
...
+            u.id, u.givenName givenName, u.surName surName, ue.email email, u.orcid orcid,
+            a.name affiliationName, a.uri affiliationRORId, a.homepage affiliationURL
+        ${fromWhereClause}
+        ORDER BY cursorId ${sortDir}
+        LIMIT ${limit + 1};
+      `;
+
+      // Prepare the values for the SQL query placeholders
+      const values = [
+        ...(context.token?.affiliationId ? [context.token?.affiliationId] : []),
```
jupiter007, line 516: L516: Are you sure the "context.token?.affiliationId" should be in an array? It looks like from `a.uri = ?` that it should be a string.

Also, it looks like if `context.token?.affiliationId` was undefined, we would have `...[]` which would resolve to being empty, and then `a.uri = ?` would get the next `%term.toLowerCase()}%` instead

src/schemas/collaborator.ts
```diff
@@ -9,7 +9,7 @@ export const typeDefs = gql`
     projectCollaborators(projectId: Int!): [ProjectCollaborator]
 
     "Search for a User to add as a collaborator"
-    findCollaborator(term: String): [CollaboratorSearchResult]
+    findCollaborator(term: String, options: PaginationOptions): CollaboratorSearchResults
```
jupiter007, line 12: It looks like `term` should be required, because in the resolver you are returning a `Bad Request` error if it isn't at least 4 characters long

briri: Yes definitely!

briri:
thanks for reviewing @jupiter007. I have made changes for the issues you pointed out

src/models/Collaborator.ts
```diff
@@ -513,7 +513,7 @@ export class ProjectCollaborator extends Collaborator {
 
       // Prepare the values for the SQL query placeholders
       const values = [
-        ...(context.token?.affiliationId ? [context.token?.affiliationId] : []),
+        ...(context.token?.affiliationId ? context.token.affiliationId : 'NULL'),
```
jupiter007, line 516: L516: you no longer need the spread operator in front of `context.token?.affiliationId` since it is no longer an array.

jupiter007 (approved):
Thanks for making the updates @briri. I just had an additional comment on the Collaborator model.

Issue #462, opened 2025-09-30
Make totalCount constant for findPotentialCollaboratorByTerm


jupiter007:
While working on the frontend ticket to run searches for project collaborators ([#403](https://github.com/CDLUC3/dmsp_frontend_prototype/issues/403)), I noticed that the `findPotentialCollaboratorByTerm` resolver was returning different `totalCount` numbers as each consecutive request was made.

We need to update the resolver to return a constant `totalCount`.

bofstein:
Is this just about returning the correct number of team members on the [Project Overview page](https://dmptool-dev.uc3dev.cdlib.net/en-US/projects/136378), or is there anything else I should be checking?  That part looks good but from the ticket it sounds like other changes might have been made.

<img width="2283" height="1321" alt="Image" src="https://github.com/user-attachments/assets/9fdf338f-ab42-4a2b-9720-c25b9dc8dec7" />

bofstein:
Ok.  I'm happy to close then, we do have this bug still open but since that's tracked we don't need this even it is related.

https://github.com/CDLUC3/dmsp_frontend_prototype/issues/1065

PR #463, opened 2025-10-01
Made updates to work with the frontend Project member search page

jupiter007:
## Description

- Updated `findPotentialCollaboratorsByTerm` in `Collaborator` model so that a constant TotalCount is returned to the client. Previously, the totalCount was changing with every cursor request.
- Updated the `create` method in `Member` class because if an `email` was not provided, then it was still return a `current` value, but just not for the member being added.
- Fixed `search` function in `User` model which was broken because of the use of `u.email` instead of `ue.email` which was breaking search.
- Updated `findCollaborator` resolver for when a user `isSuperAdmin` so that the response would be in the correct `CollaboratorSearchResults` structure.
- I updated the `CollaboratorSearchResult` schema to include `affiliationId` which is needed by client to be able to add the member.
 
Fixes # ([462](https://github.com/CDLUC3/dmsp_backend_prototype/issues/462))

## Type of change

## How Has This Been Tested?
Manual testing and existing tests

## Checklist:

### DIFF
--- a/src/resolvers/collaborator.ts
+++ b/src/resolvers/collaborator.ts
@@ -120,14 +120,8 @@ export const resolvers: Resolvers = {
             }));
 
             return {
+              ...userSearchResults,
               items: transformedItems,
-              limit: userSearchResults.limit,
-              totalCount: userSearchResults.totalCount,
-              nextCursor: userSearchResults.nextCursor,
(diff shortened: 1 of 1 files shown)


briri (approved):
Looks good @jupiter007 I just had one minor comment

Issue #251, opened 2025-04-08
Add search logic to add new Project Members


jupiter007:
We will set the search logic up so that it searches the list of existing collaborators, other people in the user's organization, and then ORCID. We will display the results in that order as well (e.g. showing an existing collaborator at the top and then any users at their org that match, and finally ORCID matches).

On the project members page we can do a match on email or ORCID. If there is a collaborator with the same value, we display the access level.

Current wireframe for Project Members search page: https://www.figma.com/proto/RRtoRLyr5Spf0KiyRfEclt/DMP-Tool-Wireframes?node-id=183-522&t=M9JzcqhvxBJdTW3Q-0&scaling=min-zoom&content-scaling=fixed&page-id=177%3A504&starting-point-node-id=177%3A642

jupiter007:
For this page, it was suggested that we search existing collaborators and other people in the user’s org, and then ORCID from our db, and then later query ORCID API. We will display results in that order as well (e.g., showing an existing collaborator at the top and then any users at their org that match, and finally ORCiD matches).

@fraserclark we may need to update the wireframe based on what you discussed in Slack to add more info about the user, such as email addresses, etc., like OPIDoR.

bofstein:
Ideal is to search:
- Current or former collaborators on any DMPs from this person
- Users in org
- ORCID database

Dependencies:
- Create data source for ORCID API (new ticket)
- Create the resolver for this function (searching members)

Note this is a one-time pull of information, not a dynamic link.  E.g. if a person changes their affiliation in ORCID a week after being added to a project, we do not expect to get that update here.  Updates are done manually by DMP owners/editors once they are pulled in

bofstein:
I'm not sure if all of these things are in yet, but I'm seeing issues with searching.  It doesn't seem to be looking up in ORCID, as I searched for various known names with 0 results.  

It also doesn't seem to work putting in a full name.  For example if I search "Brian" I find @briri  but if I search "Brian Riley" I get nothing.  And searching ORCID should bring up many more Brians.

<img width="1317" height="441" alt="Image" src="https://github.com/user-attachments/assets/a8c9d72a-7465-4897-a666-23ff5edc7ecc" />

<img width="691" height="434" alt="Image" src="https://github.com/user-attachments/assets/130b7bfb-118e-47c8-8d8e-e62faa5da896" />

Is this waiting on Backend tickets or are these bugs to file?

jupiter007:
[...]

PR #481, opened 2025-10-22
Updates to backend for Project Member Search and Add page

jupiter007:
## Description
- Updated `ORCID_REGEX` in `helper.ts` because it was not working for something like `https://sandbox.orcid.org/0009-0000-9768-4477` where there is no `www` because it was still expecting the period '.'.  Also, added a new help function just for ORCID to strip out just the ORCID id: `stripORCIDIdentifierBaseURL`
- Updated `OrcidAPI` to handle `404` errors so that we return null rather than throwing an error. Before, it was throwing an error, and then findCollaborator automatically returned an `INTERNAL_SERVER_ERROR` to the frontend.
- Updated `addProjectMemberInput` schema to include `affiliationName` since we are now using a TypeAhead field for affiliations on the Project Member Search/Add page, so sometimes we only have the `affiliationName`
- Updated `addProjectMember` resolver to use either `affiliationId` or `affiliationName` to process affiliation info 

Fixes # ([251](https://github.com/CDLUC3/dmsp_backend_prototype/issues/251))

## Type of change

## How Has This Been Tested?
Tested manually and with existing unit tests

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 - Updated `OrcidAPI` to handle `404` errors so that we return null rather than throwing an error [#251]
 - Updated `addProjectMemberInput` schema to include `affiliationName` since we are using a TypeAhead field and sometimes we only get the `affiliationName` [#251]
 - Updated `addProjectMember` resolver to use either `affiliationId` or `affiliationName` to process affiliation info [#251]
+- Updated related works resolvers to return aggregated counts for the UI, and modifications to some fields in database for UI.
 - Updated the `findCollaborator` resolver to allow for searching by ORCID or by a search term. 
 - Updated cursor pagination logic to accept sort information
 - Update `user.register` model function to auto-accept all open template and project collaboration invites
--- a/data-migrations/2025-10-16-1653-update-related-works.sql
+++ b/data-migrations/2025-10-16-1653-update-related-works.sql
@@ -1,0 +1,203 @@
+-- Add modified and modifiedById to workVersions table
+ALTER TABLE `workVersions`
+  RENAME COLUMN publishedDate TO publicationDate,
+  MODIFY COLUMN abstractText MEDIUMTEXT,
+  MODIFY COLUMN publicationVenue VARCHAR(1000);
+
+-- Procedure to create staging tables for related works
+DELIMITER $$
+
+DROP PROCEDURE IF EXISTS create_related_works_staging_tables;
+CREATE PROCEDURE `create_related_works_staging_tables`()
+BEGIN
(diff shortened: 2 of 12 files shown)


src/datasources/OrcidAPI.ts
```diff
@@ -90,9 +90,10 @@ export class OrcidAPI extends RESTDataSource {
     reference = 'OrcidAPI.getPerson'
   ): Promise<OrcidPerson | null> {
     try {
-      const path = `${OrcidConfig.apiPath}${orcid}`;
+    // Remove leading slash from path if present
```
jupiter007: We were adding too many slashes to the ORCID API url, and it was looking like `https://pub.sandbox.orcid.org//v3.0/1234-1234-1234-1234"`. So I removed the leading `/` and also removed the `/` from between the `${this.baseURL}/${path}` in the logger line.

src/datasources/OrcidAPI.ts
```diff
@@ -121,6 +122,12 @@ export class OrcidAPI extends RESTDataSource {
       );
       return null;
     } catch(err) {
+      // Handle 404 responses gracefully - person not found in ORCID
```
jupiter007: When no person was found, the try-catch for `.get` method caught a 404 error, which then threw an error, which when caught by `findCollaborator` query returned an INTERNAL_SERVER_ERROR. So, instead we return `null`

src/models/Collaborator.ts
```diff
@@ -407,7 +408,7 @@ export class ProjectCollaborator extends Collaborator {
     // Get the fully formatted ORCID
     const fullOrcid = formatORCID(orcid);
     // Get the ORCID without the base URL
-    const orcidId = stripIdentifierBaseURL(orcid);
+    const orcidId = stripORCIDIdentifierBaseURL(orcid);
```
jupiter007: L411: Previously this line was using `stripIdentifierBaseURL`, but since it didn't match any of the three base URLS, ` generalConfig.dmpIdBaseURL,  generalConfig.orcidBaseURL,generalConfig.rorBaseURL`, it didn't strip out anything and the check would fail.

So, I added another function that just extracts the ORCiD from any URL.

src/resolvers/member.ts
```diff
@@ -100,6 +101,35 @@ export const resolvers: Resolvers = {
             throw NotFoundError();
           }
 
+          // If an affiliationId was provided, check if it exists and create it if it doesn't
```
jupiter007: For the new `Project Member Search/Add` page, we are using the Affiliation Typeahead search, so we need to handle cases where only an `affiliationName` is provided.

briri (approved):
Looks good @jupiter007

PR #543, opened 2025-12-01
Fix issue with ORCID project member lookup

### LINKED ISSUE #516: ORCID returning 502 errors for project member search
The backend is returning a 502 error when trying to access the ORCID API.

If an error occurs, pass error about can't connect to ORCID in the "general" property

briri:
## Description

Fixes #516 

This was working at one point and then broke at some point along the way. I'm not sure if ORCID changed their auth URL at some point or not. It appears to only be applicable to their Sandbox env

- Updated the `.env.example` and ORCID datasource to have separate base URLs for their API and Authentication endpoints
- Reduce the noise being logged from the PlanVersion model. Found that the apollo server was running out of memory occasionally and suspect that the size of the common standard JSON might be causing this when the logger calls JSON stringify

## Type of change

## How Has This Been Tested?

Unit tests are passing and I am able to complete an ORCID lookup on my local using the UI

## Checklist:

### DIFF
--- a/src/models/PlanVersion.ts
+++ b/src/models/PlanVersion.ts
@@ -39,7 +39,7 @@ export const generateVersionSnapshot = async (
   const modified = latestVersion.modified;
 
   if (isNullOrUndefined(dmpId) || isNullOrUndefined(modified)) {
-    context.logger.error(prepareObjectForLogs(latestVersion), `${reference} - invalid DMP JSON`);
+    context.logger.error(prepareObjectForLogs({ dmpId: latestVersion?.dmp_id?.identifier }), `${reference} - invalid DMP JSON`);
     return "Invalid DMP Id or Modified Timestamp";
   }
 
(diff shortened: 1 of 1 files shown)


src/models/PlanVersion.ts
```diff
@@ -39,7 +39,7 @@ export const generateVersionSnapshot = async (
   const modified = latestVersion.modified;
 
   if (isNullOrUndefined(dmpId) || isNullOrUndefined(modified)) {
-    context.logger.error(prepareObjectForLogs(latestVersion), `${reference} - invalid DMP JSON`);
```
briri: Found that the apollo containers were quietly re-provisioning themselves in the ECS cluster. It seems like its when this model is doing things and I suspect it might be due to trying to log the entire common standard JSON

jupiter007: interesting

jupiter007 (approved):
LGTM

Issue #549, opened 2025-12-03
[Bug] - Fix users getting logged out sooner than expected


jupiter007:
Users were getting logged out sooner than expected. I believe one issue is that when the auth token is refreshed the first time, the expiration date of the `refreshToken` is overwritten from the original 24 hr time to whatever the `dmspt` auth token was given. We don't need to set the `dmpsr` token in this function, so I removed it.

PR #550, opened 2025-12-03
Removed the setting of  dmspr token in refreshTokenController

jupiter007:
## Description

Users were getting logged out sooner than expected. I believe one issue is that when the auth token is refreshed the first time, the expiration date of the `refreshToken` is overwritten from the original 24 hr time to whatever the `dmspt` auth token was given. We don't need to set the `dmpsr` token in this function, so I removed it.

Fixes # ([549](https://github.com/CDLUC3/dmsp_backend_prototype/issues/549))


briri (approved):
great catch @jupiter007. Thanks for investigating and fixing!

PR #553, opened 2025-12-05
Bug/fixed data migration issue and updated version of node-jws due to vulnerability

jupiter007:
## Description

- Fixed a small data-migration issue related to using the wrong createdById
- Updated `auth0/node-jws Improperly Verifies HMAC Signature` due to vulnerability


briri (approved):
Thanks for fixing these @jupiter007

Issue #548, opened 2025-12-03
Add seed data for project/plans


briri:
The related works data source tests are failing because they expect their to be a plan with an id of 1.

We need to add seed projects/plans anyway.

jdddog:
I fixed the related works tests in this PR: https://github.com/CDLUC3/dmsp_backend_prototype/pull/559/files#diff-440bc6410aa3ba6a8b645d90b85a3ae5781870dac56abca21bd48d791445f696

PR #575, opened 2026-01-15
Calculate summary stats for related works

jdddog:
## Description

Adds an endpoint to return summary stats about the related works associated with a plan.

## Type of change

## How Has This Been Tested?
Manually tested.

## Checklist:

### DIFF
--- a/src/models/__tests__/RelatedWork.spec.ts
+++ b/src/models/__tests__/RelatedWork.spec.ts
@@ -608,6 +608,31 @@ describe("RelatedWork queries", () => {
     const result = await RelatedWork.findByPlanAndWorkVersionId("testing", context, planId, workVersionId);
     expect(result).toEqual(null);
   });
+
+  it("statsByPlanId should call query with correct params and return the object", async () => {
+    const rw = new RelatedWork({ planId: casual.integer(1, 999), ...getMockWork() });
+    localQuery.mockResolvedValueOnce([rw]);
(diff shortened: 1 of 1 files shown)


src/models/RelatedWork.ts
```diff
@@ -400,6 +401,22 @@ export class RelatedWork extends MySqlModel {
     console.log(result);
     return Array.isArray(result) && result.length > 0 ? new RelatedWork(result[0]) : null;
   }
+
+  // Calculate related works stats for a plan
```
jupiter007: I think you should add a test for this in `RelatedWork.spec.ts`

jupiter007 (approved):
LGTM. I think you should probably add a unit test for `statsByPlanId` though

briri:
> I'm wondering if we should add a handful of data to the `relatedWorks` tables in the data migration scripts so that we can easily see the functionality when developing.

Yeah, that would be a good idea, but @jdddog won't be able to do that until we address #548. I updated that ticket to include related work data

jdddog:
> > I'm wondering if we should add a handful of data to the `relatedWorks` tables in the data migration scripts so that we can easily see the functionality when developing.
> 
> Yeah, that would be a good idea, but @jdddog won't be able to do that until we address #548. I updated that ticket to include related work data

Yeah that sounds good. I fixed the issue with the related works tests in this PR: https://github.com/CDLUC3/dmsp_backend_prototype/pull/559

PR #620, opened 2026-02-27
Fix pagination issue

jupiter007:
## Description

- Added `COALESCE` to the `cursorField` calculation so that `NULL` isn't returned, and therefore `nextCursor = NULL` any time a cursorField value was NULL.

Fixes # ([107](https://github.com/CDLUC3/dmptool-doc/issues/107))

## Type of change

## How Has This Been Tested?
Manually tested with the `template/customizations` page, which is where it was observed to be breaking.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,7 @@
 - Removed `ioredis` package
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


briri (approved):
Ah, good catch

PR #643, opened 2026-03-18
Added override for fast-sml-parser

briri:
- Added override for the `fast-xml-parser` dependency
- Updated `aws-sdk` and `@dmptool` dependencies
- Removed old overrides for `minimatch` and `immutable` as they are no longer necessary
- Removed old dependabot config since we use Renovate now


jupiter007 (approved):
LGTM

PR #646, opened 2026-03-20
Uncomment out linter/tests/audit in buildspec

briri:
Neglected to uncomment the audit, linter, and tests in the codebuild.

Also removed some debug from the data-migration process script


jupiter007 (approved):
Looks good. I noticed that when reviewing but I thought it was intentional for now. haha.

PR #647, opened 2026-03-20
added override for flatted and updated override for fast-xml-parser

briri:
Found that the build was breaking due to `fast-xml-parser`.

- Updated `fast-xml-parser` override
- Added override for `eslint` dependency `flatted`


jupiter007 (approved):
LGTM

PR #656, opened 2026-03-24
Updated VersionedTemplateSearchResult search function to return versionedTemplateCustomizationId

jupiter007:
## Description

- Updated the `VersionedTemplateSearchResult` to return `versionedTemplateCustomizationId` for the `search` function

Fixes # ([166](https://github.com/CDLUC3/dmptool-doc/issues/166))

## Type of change

## How Has This Been Tested?
Manually tested and updated unit test for the updated sql query

## Checklist:

### DIFF
--- a/src/models/VersionedTemplate.ts
+++ b/src/models/VersionedTemplate.ts
@@ -65,6 +65,7 @@ export class VersionedTemplateSearchResult {
     term: string,
     options: TemplateQueryOptions = VersionedTemplate.getDefaultPaginationOptions(),
   ): Promise<PaginatedQueryResults<VersionedTemplateSearchResult>> {
+    const userAffiliationId = context.token?.affiliationId;
     const whereFilters = ['vt.active = 1 AND vt.versionType = ?'];
     const values = [TemplateVersionType.PUBLISHED.toString()];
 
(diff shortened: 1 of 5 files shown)


src/models/VersionedTemplate.ts
```diff
@@ -103,13 +105,17 @@ export class VersionedTemplateSearchResult {
...
+    const sqlStatement = [
+      'SELECT vt.id, vt.templateId, vt.name, vt.description, vt.version, vt.visibility, vt.bestPractice,',
+      'vt.modified, vt.modifiedById, TRIM(CONCAT(u.givenName, " ", u.surName)) as modifiedByName,',
+      'a.id as ownerId, vt.ownerId as ownerURI, a.displayName as ownerDisplayName,',
+      'a.searchName as ownerSearchName,',
+      'vtc.id as versionedTemplateCustomizationId',
+      'FROM versionedTemplates vt',
+      'LEFT JOIN users u ON u.id = vt.modifiedById',
+      'LEFT JOIN affiliations a ON a.uri = vt.ownerId',
+      'LEFT JOIN versionedTemplateCustomizations vtc ON vtc.currentVersionedTemplateId = vt.id'
```
briri: I think we want to limit this down to the current user's affiliation otherwise its going to return a record for every affiliation that has customized the template.

See if adding an additional argument of `AND vtc.affiliationId = ${context.token.affiliationId}` to this `LEFT JOIN` will work.

It should give you the correct `versionedTemplateCustomizationId` for the user or NULL if they haven't customized it yet.

briri (changes requested):
I think you need to adjust the query a bit. The new join will return a record for every affiliation that has customized the template.

src/models/VersionedTemplate.ts
jupiter007: Oh yes! Thanks @briri. Good catch.

I updated the query to check that vtc.affiliationId matches user's affiliation.

Thank you

briri (approved):
Thanks for updating @jupiter007

PR #666, opened 2026-03-29
Feature/167/update plan query to return customization data

jupiter007:
## Description

- Updated `PlanSectionProgress` so that it returns custom sections, and correct totalQuestions counts that include customQuestions
- Updated `PlanSearchResult` `versionedSections` chained resolver to include `versionedTemplateId`

Fixes # ([167](https://github.com/CDLUC3/dmptool-doc/issues/167))

## Type of change

## How Has This Been Tested?
Manually tested with Apollo Explorer, also with frontend branch, https://github.com/CDLUC3/dmptool-ui/pull/1217, and with new unit tests

## Checklist:


src/models/__mocks__/Plan.ts
```diff
@@ -67,6 +67,7 @@ const planToPlanSearchResult = (plan: Plan): PlanSearchResult => {
     funding: casual.company_name,
     members: casual.full_name,
     templateTitle: casual.title,
+    versionedTemplateId: plan.versionedTemplateId,
```
jupiter007, line 70: Added `versionedTemplateId` to `PlanSearchResult` so that we can include the `versionedCustomSections` in the `versionedSections` chained resolver for `plan`

src/models/CustomQuestion.ts
```diff
@@ -324,4 +324,32 @@ export class CustomQuestion extends MySqlModel {
...
+  /**
+ * Find all custom questions for a specific section within a template customization
+ *
+ * @param reference The reference to use for logging errors.
+ * @param context The Apollo context.
+ * @param templateCustomizationId The id of the template customization.
+ * @param sectionId The id of the section (either a versionedSection id or customSection id).
+ * @returns The custom questions.
+ */
+  static async findByCustomizationAndSectionId(
```
jupiter007, line 337: This was added as part of an update to TemplateCustomization.ts because the `publish` mutation was not saving to `versionedCustomSections` or `versionedCustomQuestions`

src/models/Plan.ts
```diff
@@ -108,21 +110,31 @@ export class PlanSearchResult {
...
+  CUSTOM = 'CUSTOM',
+}
+
+
 /**
  * Class that represents the progress of a plan section.
  * This includes the total number of questions and the percentage of questions
  * answered across all sections of the template.
  */
 export class PlanSectionProgress {
```
jupiter007, line 124: Made a lot of updates to `PlanSectionProgress` so that the `findByPlanId` called by the `versionedSections` chained resolver inthe `Plan` model would include both `BASE` and `CUSTOM` sections and correct `totalQuestion` counts

src/models/TemplateCustomization.ts
```diff
@@ -651,6 +655,65 @@ export class TemplateCustomization extends MySqlModel {
         const created: VersionedTemplateCustomization = await newVersion.create(context);
 
         if (!isNullOrUndefined(created) && !created.hasErrors() && created.id) {
+          // Snapshot custom sections into versionedCustomSections
```
jupiter007, line 658: This is a fix to update `versionedCustomSection` and `versionedCustomQuestions` table when a templateCustomization is published

src/resolvers/feedback.ts
```diff
@@ -112,11 +112,8 @@ export const resolvers: Resolvers = {
             throw NotFoundError(`Plan with ID ${planId} not found`);
           }
 
-          // Get versionedTemplate associated with the plan
```
jupiter007, line 115: The `&& context.token.affiliationId === versionedTemplate.ownerId` condition was actually breaking the page. If a CDL user, who used an NSF template to create a plan, tried to come back and edit the plan, a "Forbidden" error was displaying at the top because of this `feedback` condition, so i removed it, because it didn't make sense

src/resolvers/plan.ts
```diff
@@ -326,7 +327,7 @@ export const resolvers: Resolvers = {
     },
     versionedSections: async (parent: Plan, _, context: MyContext): Promise<PlanSectionProgress[]> => {
       if (parent?.id) {
-        return await PlanSectionProgress.findByPlanId('plan versionedSections resolver', context, parent.id);
+        return await PlanSectionProgress.findByPlanId('plan versionedSections resolver', context, parent.id, parent?.versionedTemplateId);
```
jupiter007, line 330: added the `versionedTemplateId` to these chained resolvers because we need it to get the versionedCustomSections

src/resolvers/project.ts
```diff
@@ -51,8 +51,8 @@ export const resolvers: Resolvers = {
       try {
         if (isAuthorized(context.token)) {
           const pagOpts = !isNullOrUndefined(paginationOptions) && paginationOptions.type === PaginationType.OFFSET
-                      ? paginationOptions as PaginationOptionsForOffsets
-                      : { ...paginationOptions, type: PaginationType.CURSOR } as PaginationOptionsForCursors;
+            ? paginationOptions as PaginationOptionsForOffsets
```
jupiter007, line 54: Sorry, just automated formatting due to Prettier

src/models/CustomQuestion.ts
briri: good catch!

src/models/Plan.ts
```diff
@@ -131,14 +143,93 @@ export class PlanSectionProgress {
...
+  */
+  private static async findTemplateCustomizationId(
+    reference: string,
+    context: MyContext,
+    versionedTemplateId: number
+  ): Promise<number | undefined> {
+    const sql = `
+      SELECT templateCustomizationId
+      FROM versionedTemplateCustomizations
+      WHERE currentVersionedTemplateId = ?
```
briri, line 158: A template can and will have many customizations, so you likely want to pass in the affiliationId as well. I don't think always grabbing the first result will work here.

For example Berkeley customizes the NSF template and then Stanford customizes the NSF template. Which one would you return?

Depending on the context: 
- If the plan is being created we should use the the affiliation of the current user.
- If the plan is being updated or viewed we should use the affiliation of creator

src/models/Plan.ts
```diff
@@ -177,14 +268,91 @@ export class PlanSectionProgress {
...
-        return new PlanSectionProgress(entry);
+        return new PlanSectionProgress({ ...entry, sectionType: PlanSectionType.BASE });
       })
       : [];
+
+    // If there are no base sections the plan is in a bad state — return early
+    if (!baseSections.length) return baseSections;
+
+    // If versionedTemplateId wasn't provided, skip customization lookup
+    if (isNullOrUndefined(versionedTemplateId)) return baseSections;
```
briri: Is this possible? A Plan should always have a versionedTemplateId

src/resolvers/feedback.ts
briri: Update this to validate that the ADMIN belongs to the same organization as the one who owns the plan (same affiliation as the creator).

briri: otherwise any admin would be be able to provide feedback on any plan

briri (changes requested):

src/models/Plan.ts
jupiter007: Ah right. I will add the affiliationId to the query

src/models/Plan.ts
jupiter007: Oh I guess not. Just checked and versionedTemplateId is `NOT NULL` so I can remove that line. Thanks

src/resolvers/feedback.ts
jupiter007: Thanks Brian. I've updated it to check if user's affiliationId matches the plan creator's affiliationId.

jupiter007:
Thanks so much for your review @briri. I've responded to your comments. Please take another look when you get the chance.

briri (approved):
Thanks for making those changes. Looks good

PR #669, opened 2026-03-31
Bug fix for Answer default and also update dependencies

briri:
- Since the research output table no longer requires the user to define an output type, but the RDA commmon standard requires it, updated the Answer so that it sets the default output type for each row in the research output table
- Updates to dependencies.
- Also updated the name and description of the application in the `package.json`


src/models/Answer.ts
```diff
@@ -68,12 +75,63 @@ export class Answer extends MySqlModel {
   prepForSave(): void {
     // Remove leading/trailing blank spaces
     this.json = this.json?.trim();
+
+    const parsedJSON = JSON.parse(this.json);
```
briri, line 79: I thought I had submitted the PR for this before I left, but apparently not. This just ensures that every row in a research output table has an output type.

jupiter007 (approved):
LGTM

PR #682, opened 2026-04-09
Updates for custom question answers

jupiter007:
## Description

- Added `fetchAnswerCustomQuestion` to `Plan` model in order to get answered counts for custom questions
- Updated `PlanProgress.findByPlanId` to use `fetchAnswerCustomQuestions` and return answered question counts that include the custom questions
- Updated `PlanProgress.findByPlanId` to reuse the `PlanSectionProgress.findByPlanId` function to return `totalQuestions` and `answeredQuestions`

Fixes # ([161](https://github.com/CDLUC3/dmptool-doc/issues/161))

## Type of change

## How Has This Been Tested?
Manually tested and tested with updated unit tests

## Checklist:

### DIFF
--- a/src/models/Plan.ts
+++ b/src/models/Plan.ts
@@ -224,6 +224,44 @@ export class PlanSectionProgress {
   }
 
   /**
+   * Fetch the count of custom questions that have been answered by section type for a given plan and template customization. 
+   * This allows us to credit answered custom questions in the progress calculation for both base and custom sections.
+   * @param reference
+   * @param context 
+   * @param planId 
+   * @param templateCustomizationId 
+   * @returns 
+   */
+  private static async fetchAnsweredCustomQuestions(
+    reference: string,
+    context: MyContext,
+    planId: number,
+    templateCustomizationId: number,
+  ): Promise<{ sectionId: number; sectionType: string; answeredCount: number }[]> {
+    const sql = `
+    SELECT
+      vcq.versionedSectionId  AS sectionId,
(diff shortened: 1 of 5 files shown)


src/models/Plan.ts
```diff
@@ -223,6 +223,44 @@ export class PlanSectionProgress {
...
+      vcq.versionedSectionType AS sectionType,
+      COUNT(DISTINCT a.versionedCustomQuestionId) AS answeredCount
+    FROM answers a
+    JOIN versionedCustomQuestions vcq
+      ON vcq.id = a.versionedCustomQuestionId
+    JOIN versionedTemplateCustomizations vtc
+      ON vtc.id = vcq.versionedTemplateCustomizationId
+    WHERE a.planId = ?
+      AND vtc.templateCustomizationId = ?
+      AND NULLIF(TRIM(a.json), '') IS NOT NULL
```
briri: Not sure if this is what you want. That's a JSON column, which would not accept a series of spaces because its invalid JSON. So there's no need to do the `TRIM`, that forces MySQL to cast it as a string.

Try one of MySQL's JSON functions like `JSON_TYPE(a.json) = 'OBJECT'` instead

briri (approved):
Looks good @jupiter007 I just had one refinement to help optimize one of your queries.

src/models/Plan.ts
jupiter007: Makes sense. I've made the change and will merge.

PR #689, opened 2026-04-15
Fix/re3data script

briri:
Found some typescript errors when investigating how to load the re3data into the AWS environments. This PR addresses those issues and updates some of the documentation


jupiter007 (approved):
LGTM

PR #118, opened 2024-10-04
Feature/versioning

briri:
## Description

Please include a summary of the changes and the related issue. Are there any dependencies related to this change?

Fixes #108

Add versioning logic for Templates, Sections and Questions
- Updated data-migrations for QuestionType to include an isDefault boolean flag
- Added QuestionService
- Added Question, QuestionType and QuestionCondition models (as much as needed to generate versions)
- Added VersionedQuestion and VersionedQuestionCondition models
- Updated SectionService to include versioning logic
- Update TemplateService to cascade versioning to Sections and Questions
- Added tests for new models
- Started to add tests for services, but did not finish

## Type of change
Please delete options that are not relevant

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Include any relevant details for your test configuration.

I added some new tests for the models. Existing tests pass still.

Started to add tests for the services but did not finish

## Checklist:
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I updated the CHANGELOG.md and added documentation if necessary
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #704, opened 2026-04-22
Update to have re3data repository searches run against AOSS

briri:
## Description

Fixes issue described in [#118](https://github.com/CDLUC3/dmptool-doc/issues/118) with the OpenSearch index being unavailable.

- Added new `OPENSEARCH_SERVERLESS_NODE` environment variable
- Refactored `re3data-os-populate.ts` to use `fast-sml-parser` instead of `@xmldom/xmldom` which is slower
- Updated `re3data-os-populate.ts` to write to an AWS Serverless OpenSearch (AOSS) instance 
- Updated `OpenSearch` datasource to include a new `serverlessClient` property and `createOpenSearchServerlessClient` function.
- Also addressed an issue where the Apollo server logs were not properly piping to the CloudWatch logs in the new combined ApolloShib ECS tasks.

## Type of change

## How Has This Been Tested?

The server uses the same OS instance as the related works matching processes when running locally. When running in the AWS environment, the related works matching continue to come form the OS domain but the re3data repositories are now stored in the AOSS collection. I have verified in the CloudWatch logs that the Apollo server is calling AOSS and returning results.

## Checklist:


data-migrations/dataSync/re3data-os-populate.ts
```diff
@@ -38,390 +41,168 @@ const OPENSEARCH_CONFIG = {
...
-const getAllText = (node: Node, path: string): string[] => {
-  const select = xpath.useNamespaces(ns);
-  try {
-    const results = select(path, node) as Node[];
-    return results.map(n => n.textContent || '').filter(s => s.trim() !== '');
-  } catch {
-    return [];
-  }
-};
+const parser = new XMLParser({
```
briri, line 50: fast-xml-parser is supposed to be much faster than the xml-dom package, so I refactored to use that one

src/resolvers/__tests__/collaborator.spec.ts
```diff
@@ -16,9 +16,6 @@ jest.mock('../../context.ts');
 jest.mock('../../datasources/cache');
 jest.mock('../../services/emailService');
 jest.mock('../../services/openSearchService');
-jest.mock('../../config/awsConfig', () => ({
```
briri, line 19: I realized that these were redundant since the awsConfig is mocked in the `__tests__/setup.ts`

src/services/openSearchService.ts
```diff
@@ -272,6 +279,11 @@ export class OpenSearchService {
               uri: uris,
             },
           },
+          sort: {
```
briri, line 282: this was an attempt to get the results to sort. They're still showing up in the UI as descending. I'm not sure if this just isn't working, or if its being resorted elsewhere before getting displayed

src/logger.ts
```diff
@@ -26,14 +26,21 @@ export const REDACTION_MESSAGE = '[MASKED]';
...
   },
   ...ecsFormat()
-});
+};
+
+export const logger: Logger = process.env.NODE_ENV === 'production'
+  // In the AWS environment, this application runs on a container with other processes,
+  // so we need to redirect the logs so they make it to the correct place so supervisord
+  // can pick them up. We also need to run synchronously
+  ? pino(basePinoConfig, pino.destination({ sync: true, dest: 1 }))
```
briri, line 42: I found that the logs weren't making into Cloudwatch/OpenSearch and it was because of the way the new combined Apollo server + Shibboleth + Apache HTTPD container was configured. 

This tells Pino to write logs synchronously and to the supervisord's process

data-migrations/dataSync/re3data-os-populate.ts
```diff
@@ -38,390 +41,168 @@ const OPENSEARCH_CONFIG = {
...
+        const detailXml: string = await detailRes.text();
+        const detailObj = parser.parse(detailXml) as Record<string, unknown>;
+        const re3data = detailObj.re3data as Record<string, unknown> | undefined;
+        const r = re3data?.repository as Record<string, unknown> | undefined;
+
+        if (!r) {
+          // If the root isn't 're3data', try just 'repository' (sometimes happens
+          // with fast-xml-parser NSPrefix removal)
+          const fallbackR: unknown = detailObj.repository;
+          if (!fallbackR) {
```
jupiter007: L132: In the case that `r` is undefined, could `fallbackR`exist? If so, would we want to assign `fallbackR` to `r`?

data-migrations/dataSync/re3data-os-populate.ts
```diff
@@ -38,390 +41,168 @@ const OPENSEARCH_CONFIG = {
...
-
-          created: new Date().toISOString(),
-          modified: new Date().toISOString()
+          repositoryTypes: ensureArray(r.type).map((t: unknown): string => getVal(t)),
+          subjects: ensureArray(r.subject).map((s: unknown): string =>
+            getVal(s).replace(/^\s*\d+\s*/, '').trim()
+          ),
+          provider_types: ensureArray(r.providerType).map((k: unknown): string => getVal(k)),
+          keywords: ensureArray(r.keyword).map((k: unknown): string => getVal(k)),
+          access: ensureArray(r.databaseAccess).map((a: Record<string, unknown>): string => {
```
jupiter007: L151: `access` is assigned as a array of strings here, but in the `Re3DataRepository` type definition it is a string: `  /** Access restrictions */
  access?: Maybe<Scalars['String']['output']>;`. Also, in the previous code, it was assigned as a string.

data-migrations/dataSync/re3data-os-populate.ts
```diff
@@ -38,390 +41,168 @@ const OPENSEARCH_CONFIG = {
...
+          upload_types: ensureArray(r.dataUpload).map((u: Record<string, unknown>): string => getVal(u.dataUploadType)),
+          certificates: ensureArray(r.certificate).map((c: unknown): string => getVal(c)),
+          software: ensureArray(r.software).map((s: Record<string, unknown>): string => getVal(s.softwareName)),
+          created: getVal(r.entryDate) as string,
+          modified: getVal(r.lastUpdate) as string,
+          synDate: new Date().toISOString(),
         };
 
-        if (VERBOSE) console.log('Prepared document:', repositoryData.id, repositoryData.name || '(no name)');
+        console.log(`Indexing ${id} - ${doc.name} ...`);
```
jupiter007, line 189: Thank you for adding this console.log()! It's reassuring and helpful to see the data that is being synced.

data-migrations/dataSync/re3data-os-populate.ts
```diff
@@ -38,390 +41,168 @@ const OPENSEARCH_CONFIG = {
...
-              errored.slice(0, 10).forEach(e => {
-                const op = Object.values(e.it as Record<string, unknown>)[0] as Record<string, unknown> | undefined;
-                console.error('Item error:', { id: op ? (op['_id'] ?? null) : null, error: op ? (op['error'] ?? null) : null });
-              });
-            } else if (VERBOSE) {
-              console.log('Bulk index successful for this batch');
-            }
-
-            currentBatch = [];
+        if (currentBatch.length / 2 >= argv['batch-size'] || i === repoIds.length - 1) {
```
jupiter007: L171: If `argv['batch-size']` is `undefined`, JavaScript will set this to `false` and then it will fall back to to `i === repoIds.length -1` which will only flush on the last repository in the list. We can use what was being used before in place of that check, which is `OPENSEARCH_CONFIG.batchSize`which guarantees that `batchSize` has a number.

data-migrations/dataSync/re3data-os-populate.ts
```diff
@@ -38,390 +41,168 @@ const OPENSEARCH_CONFIG = {
...
-      console.log(`Alias ${ALIAS_NAME} now points to ${newIndexName}. Old indices cleaned up.`);
-    } catch (err) {
-      console.error('Failed to swap alias or cleanup old indices:', err instanceof Error ? err.message : err);
-      throw err;
+    const actions: unknown[] = [{ add: { index: newIndexName, alias: ALIAS_NAME } }];
+    if (aliasExists) {
+      actions.unshift({ remove: {index: `${OPENSEARCH_CONFIG.indexPrefix}-*`, alias: ALIAS_NAME } });
     }
 
+    await client.indices.updateAliases({ body: { actions } });
```
jupiter007, line 223: Do we have to delete the old indices after updating aliases like in the old code? ```// cleanup old indices that match prefix and are not the new index
const existing = await listIndicesWithPrefix(prefix);
const toDelete = existing.filter(n => n !== newIndex);
for (const idx of toDelete) {
  try {
    if (VERBOSE) console.log(`Deleting old index ${idx}...`);
    const resp = await client.indices.delete({ index: idx });
    if (VERBOSE) console.log('Delete response:', ((resp as unknown) as { body?: unknown }).body || resp);
  } catch (err) {
    console.warn(`Failed to delete old index ${idx}:`, err instanceof Error ? err.message : err);
  }
}```

data-migrations/dataSync/re3data-os-populate.ts
```diff
@@ -38,390 +41,168 @@ const OPENSEARCH_CONFIG = {
 const client = new Client({
```
jupiter007, line 41: L41: Could the new `createOpenSearchServerlessClient` function be used here?

src/resolvers/__tests__/collaborator.spec.ts
jupiter007: Nice

src/logger.ts
jupiter007: Good catch!

data-migrations/dataSync/re3data-os-populate.ts
```diff
@@ -38,390 +41,168 @@ const OPENSEARCH_CONFIG = {
...
-    }
+    // Get Repo List
+    const listRes = await fetch(`${RE3DATA_API_BASE}/repositories`);
+    const listXml = await listRes.text();
+    const listObj = parser.parse(listXml) as Record<string, unknown>;
+    const list = listObj.list as Record<string, unknown> | undefined;
+    const listRepo = list.repository as Record<string, unknown>[] | undefined;
+
+    // Adjusting based on re3data API structure
+    let repoIds = ensureArray(listRepo).map((r: Record<string, unknown>)  => r.id);
```
jupiter007, line 113: I'm seeing type errors here. For instance on L87, it says that `list` is possibly null, so you could add "optional chaining" here. Also, on L90 it says ```Argument of type '(r: Record<string, unknown>) => unknown' is not assignable to parameter of type '(value: unknown, index: number, array: unknown[]) => unknown'.
  Types of parameters 'r' and 'value' are incompatible.
    Type 'unknown' is not assignable to type 'Record<string, unknown>'.ts(2345)```

jupiter007: I double checked and I'm using the workspace Typescript. Are you able to see these type errors in Webstorm?

jupiter007 (commented):
Looks good. I tested it and it was definitely faster to load the data than before. Thanks for working on this.

I just had a couple of comments.

package.json
```diff
@@ -58,7 +58,6 @@
     "@keyv/redis": "^5.1.6",
     "@node-oauth/express-oauth-server": "^4.1.5",
     "@opensearch-project/opensearch": "^3.5.1",
-    "@xmldom/xmldom": "^0.9.9",
```
jupiter007: Convenient that we could get rid of this, since it has security vulnerabilities 😄

data-migrations/dataSync/re3data-os-populate.ts
briri: Thanks. I will update

data-migrations/dataSync/re3data-os-populate.ts
briri: Yeah. I noticed that after running a few tests

data-migrations/dataSync/re3data-os-populate.ts
briri: Yeah probably. I'd rather not refactor now though. We can do this when we build the next sync for the metadata standards, licenses or ROR data. We're going to want to rename this to re3dataSync at that point, and I'm guessing between them all, there will be some of this code that we will want to move int othe shared `@dmptol/utils` repo

briri:
Made some changes based on your feedback @jupiter007

PR #711, opened 2026-04-30
Added feedback status to plan

jupiter007:
## Description

- Added feedbackStatus to the plan query, so the client can know whether to display feedback notification at top of question page
- Fixed bug in `openSearchService` that was throwing an error and not returning repositories

Fixes # ([196](https://github.com/CDLUC3/dmptool-doc/issues/196))

## Type of change

## How Has This Been Tested?
Tested manually and with frontend changes here: https://github.com/CDLUC3/dmptool-ui/tree/feature/196/JS-update-textArea-question-with-feedback-mode

## Checklist:

### DIFF
--- a/src/resolvers/feedback.ts
+++ b/src/resolvers/feedback.ts
@@ -105,38 +105,32 @@ export const resolvers: Resolvers = {
     planFeedbackStatus: async (_, { planId }, context: MyContext): Promise<PlanFeedbackStatus> => {
       const reference = 'planFeedbackStatus resolver';
       try {
-        // if the user is an admin
-        if (isAdmin(context.token)) {
-          // Check to see if planId exists in our records
-          const plan = await Plan.findById(reference, context, planId);
-          if (!plan) {
-            throw NotFoundError(`Plan with ID ${planId} not found`);
-          }
-
-          if (!plan.createdById) {
-            throw NotFoundError(`Plan with ID ${planId} has no creator`);
-          }
-
-          // Fetch the plan creator to compare affiliations
-          const planCreator = await User.findById(reference, context, plan.createdById);
-          if (!planCreator) {
-            throw NotFoundError(`User who created plan ${planId} not found`);
-          }
-
-          // If the user is a superAdmin or an admin for the same affiliation
-          if (isSuperAdmin(context.token) || (isAdmin(context.token) && context.token.affiliationId === planCreator.affiliationId)) {
-
-            // Check that user has permissions to access feedback
-            const projectId = plan.projectId;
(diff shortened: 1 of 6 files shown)


src/services/openSearchService.ts
```diff
@@ -291,7 +291,7 @@ export class OpenSearchService {
             },
           },
           sort: {
-            name: {
+            "name.keyword": {
```
jupiter007, line 294: Had to update this because not including the `keyword` was causing an error to be thrown when this method was called

src/resolvers/plan.ts
```diff
@@ -319,6 +319,13 @@ export const resolvers: Resolvers = {
       }
       return [];
     },
+    feedbackStatus: async (parent: Plan, _, context: MyContext): Promise<PlanFeedbackStatus> => {
```
jupiter007, line 322: Returning feedbackStatus with `plan` query response, because it can use the logic from `PlanFeedback.statusForPlan` and makes it easier for client to determine how to handle the page functionality

src/resolvers/feedback.ts
```diff
@@ -105,38 +105,32 @@ export const resolvers: Resolvers = {
...
-          // Check to see if planId exists in our records
-          const plan = await Plan.findById(reference, context, planId);
-          if (!plan) {
-            throw NotFoundError(`Plan with ID ${planId} not found`);
-          }
-
-          if (!plan.createdById) {
-            throw NotFoundError(`Plan with ID ${planId} has no creator`);
-          }
+        // Check to see if planId exists in our records
```
jupiter007, line 108: Removed requirement that only `admin` can access the `planFeedbackStatus` since we need that info for any user on the Plan Overview page

src/resolvers/feedback.ts
```diff
@@ -105,38 +105,32 @@ export const resolvers: Resolvers = {
...
-          const planCreator = await User.findById(reference, context, plan.createdById);
-          if (!planCreator) {
-            throw NotFoundError(`User who created plan ${planId} not found`);
-          }
+        if (!plan.createdById) {
+          throw NotFoundError(`Plan with ID ${planId} has no creator`);
+        }
 
-          // If the user is a superAdmin or an admin for the same affiliation
-          if (isSuperAdmin(context.token) || (isAdmin(context.token) && context.token.affiliationId === planCreator.affiliationId)) {
```
briri, line 127: Good, thanks for cleaning this up. The `hasPermissionOnProject` covers this.

briri (approved):
LGTM

jupiter007:
> LGTM

Thanks for your review Brian

Issue #213, opened 2025-02-25
Project search backend changes


briri:
Update project schema to add search 

Add new datasources for each of the APIs

Add code to convert API results into Project, ProjectContributor models

Add resolver for search  to the projects resolver

Refer to corresponding frontend requirements: https://github.com/CDLUC3/dmsp_frontend_prototype/issues/352

PR #713, opened 2026-05-04
Add alternate identifiers to Plans

briri:
## Description

To support REST API work on [#213](https://github.com/CDLUC3/dmptool-doc/issues/213)

The RDA Common Standard requires a DMP id to be specified, but callers of the REST API will not have one when they attempt to create a DMP because its something we define internally.

The REST API will convey the caller's id in the `alternate_identifier` array within the maDMP JSON so this PR adds a new `AlternateIdentifier` model and a new resolver called `planByAlternateIdentifier`.

- Added new data migration to create an `alternateIdentifiers` table
- Added a new `AlternateIdentifier` model
- Added a new `planByAlternateIdentifier` resolver and query in schema
- Added new `addAlternateIdentifierToPlan` and `removeAlternateIdentifierFromPlan` mutations
- Added `alternateIdentifiers` as a chained resolver on `Plan`

## Type of change

## How Has This Been Tested?

Added new unit tests, and verified query works in apollo explorer, otherwise this is a completely new model and resolver and shouldn't impact anything else

## Checklist:


src/schemas/versionedTemplate.ts
```diff
@@ -13,6 +13,9 @@ export const typeDefs = gql`
 
     "Get all of the customizable templates for the current user's affiliation (user must be an Admin)"
     customizableTemplates(term: String, status: String, migrationStatus: String, paginationOptions: PaginationOptions): CustomizableTemplateSearchResults
+
+    "Get the default best practice template."
+    defaultTemplate: VersionedTemplate
```
briri, line 18: this leaked in, but I'm going to send a separate PR for the rest of it shortly. It shouldn't cause any downstream issues because its a new query

src/models/AlternateIdentifier.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    if (!this.planId) this.addError('planId', 'Plan can\'t be blank');
+    if (!validateURL(this.alternateIdentifier)) this.addError('alternateIdentifier', 'Alternate identifier can\'t be blank');
+
+    return Object.keys(this.errors).length === 0;
+  }
+
+  // Save the current record
+  async create(context: MyContext): Promise<AlternateIdentifier> {
+    // First make sure the record doesn't already exist
```
jupiter007: L31: Minor - duplicate comment with L38

src/models/AlternateIdentifier.ts
```diff
@@ -0,0 +1,80 @@
...
+      const result = await AlternateIdentifier.delete(context, AlternateIdentifier.tableName, this.id, 'AlternateIdentifier.delete');
+      if (result) {
+        return new AlternateIdentifier(this);
+      }
+    }
+    return null;
+  }
+
+  // Return the specified AlternateIdentifier
+  static async findById(reference: string, context: MyContext, id: number): Promise<AlternateIdentifier> {
```
jupiter007: L68: ways the `findById` function returns `AlternateIdentifer` but it can also be `null`

src/models/AlternateIdentifier.ts
```diff
@@ -0,0 +1,80 @@
...
+    if(await this.isValid()) {
+      if (current) {
+        const assoc = current.planId == this.planId ? 'this Plan' : 'another Plan';
+        this.addError(
+          'general',
+          `That identifier is already associated with ${assoc}. Consider using a domain namespace to make it unique.`
+        );
+      } else {
+        // Save the record and then fetch it
+        const newId = await AlternateIdentifier.insert(context, AlternateIdentifier.tableName, this, 'AlternateIdentifier.create');
```
jupiter007, line 46: L48: `newId` can be `null`, so we should probably add some sort of check like:
```
if (!newId) {
  this.addError('general', 'Failed to save the record.');
  return new AlternateIdentifier(this);
}
```

src/models/AlternateIdentifier.ts
```diff
@@ -0,0 +1,80 @@
...
+  async create(context: MyContext): Promise<AlternateIdentifier> {
+    // First make sure the record doesn't already exist
+    const current = await AlternateIdentifier.findByAlternateIdentifier(
+      'AlternateIdentifier.create',
+      context,
+      this.alternateIdentifier,
+    );
+
+    // Then make sure it doesn't already exist
+    if(await this.isValid()) {
```
jupiter007, line 30: L39: Might make sense to check for `this.isValid()` before calling `AlternateIdentifier.findByAlternateIdentifier` on L32.

src/models/AlternateIdentifier.ts
```diff
@@ -0,0 +1,80 @@
...
+      if (current) {
+        const assoc = current.planId == this.planId ? 'this Plan' : 'another Plan';
```
jupiter007: L41: This should probably be using a strict equality operator here instead of the loose `==`

src/models/AlternateIdentifier.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.planId = options.planId;
+    this.alternateIdentifier = options.alternateIdentifier;
+  }
+
+  // Validation to be used prior to saving the record
+  async isValid(): Promise<boolean> {
+    await super.isValid();
+
+    if (!this.planId) this.addError('planId', 'Plan can\'t be blank');
+    if (!validateURL(this.alternateIdentifier)) this.addError('alternateIdentifier', 'Alternate identifier can\'t be blank');
```
jupiter007: L24: Is the `alternateIdentifier` always a URL? Just checking since the `validateURL` only validates for a url. Also, the message might be updated to say `invalid identifier` instead of `can't be blank`?

src/resolvers/plan.ts
```diff
@@ -66,6 +67,37 @@ export const resolvers: Resolvers = {
...
+    // Lookup a Plan by its alternate identifier
+    planByAlternateIdentifier: async(_, { alternateIdentifier }, context: MyContext): Promise<Plan> => {
+      const reference = 'planByAlternateIdentifier resolver';
+      try {
+        const identifier: AlternateIdentifier = await AlternateIdentifier.findByAlternateIdentifier(
+          reference,
+          context,
+          alternateIdentifier
+        );
+        if (!alternateIdentifier) {
```
jupiter007: L80: Shouldn't this be checking for `identifier` and not for `alternateIdentifier`?

src/resolvers/plan.ts
```diff
@@ -66,6 +67,37 @@ export const resolvers: Resolvers = {
...
+          throw NotFoundError('Alternate identifier not found');
+        }
+
+        const plan = await Plan.findById(reference, context, identifier.planId);
+        if (!plan) {
+          throw NotFoundError(`Plan with ID not found`);
+        }
+
+        const project = await Project.findById(reference, context, plan.projectId);
```
jupiter007, line 94: L89: Looks like L89 can return a `null`, so we might want to add something like:
```
if (!project) {
  throw NotFoundError('Project not found');
}
```

src/resolvers/plan.ts
```diff
@@ -66,6 +67,37 @@ export const resolvers: Resolvers = {
...
+          context,
+          alternateIdentifier
+        );
+        if (!alternateIdentifier) {
+          throw NotFoundError('Alternate identifier not found');
+        }
+
+        const plan = await Plan.findById(reference, context, identifier.planId);
+        if (!plan) {
+          throw NotFoundError(`Plan with ID not found`);
```
jupiter007: L86: Should we include the `planId` in the Not Found error message?

jupiter007 (approved):
Thanks. LGTM.

PR #714, opened 2026-05-05
Updated completeFeedback 

jupiter007:
## Description

- Updated `completedFeedback` resolver to send an email to the plan owner when feedback is marked as complete, and added a `sendEmail` boolean flag to input variables to stop an email from being sent
- Added `planOwner` field to the `plan` query response so that client can get plan owner affiliation.uri for feedback pages
- Added `feedback.spec.ts` unit tests for the feedback resolver

Fixes # ([331](https://github.com/CDLUC3/dmptool-apollo-server/issues/331))([198](https://github.com/CDLUC3/dmptool-doc/issues/198))

## Type of change

## How Has This Been Tested?
Manually tested and added unit tests

## Checklist:

### DIFF
--- a/src/resolvers/feedback.ts
+++ b/src/resolvers/feedback.ts
@@ -226,7 +226,6 @@ export const resolvers: Resolvers = {
               throw NotFoundError(`PlanFeedback with ID ${planFeedbackId} not found`);
             }
 
-
             const newFeedback = new PlanFeedback({
               id: feedback.id,
               planId: feedback.planId,
@@ -243,13 +242,13 @@ export const resolvers: Resolvers = {
             }
 
             if (sendEmail) {
-              const planURL = `/projects/${project.id}/plans/${planId}`;
+              const planURL = `/projects/${project.id}/dmp/${planId}`;
               const adminName = [context.token.givenName, context.token.surName]
                 .filter(Boolean).join(' ');
 
               await sendFeedbackCompleteEmail(
                 context,
-                plan.createdById,
+                feedback.requestedById,
                 adminName,
                 plan.title,
                 planURL,
--- a/src/resolvers/feedback.ts
+++ b/src/resolvers/feedback.ts
@@ -10,7 +10,11 @@ import {
 
 import { isAuthorized } from "../services/authService";
 import { hasPermissionOnProject } from "../services/projectService";
-import { sendProjectCollaboratorsCommentsAddedEmail, sendFeedbackRequestEmail } from '../services/emailService';
+import {
+  sendProjectCollaboratorsCommentsAddedEmail,
+  sendFeedbackRequestEmail,
+  sendFeedbackCompleteEmail
+} from '../services/emailService';
 import { isAdmin, isSuperAdmin } from "../services/authService";
 import { canDeleteComment } from "../services/commentPermissions";
 import { Project } from "../models/Project";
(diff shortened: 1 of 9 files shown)


src/resolvers/feedback.ts
```diff
@@ -237,6 +241,21 @@ export const resolvers: Resolvers = {
             if (!updatedFeedback) {
               throw NotFoundError(`PlanFeedback with ID ${planFeedbackId} not found`);
             }
+
+            if (sendEmail) {
+              const planURL = `/projects/${project.id}/plans/${planId}`;
```
briri: shouldn't this be `dmp/${planId}`?

src/resolvers/plan.ts
```diff
@@ -284,6 +285,13 @@ export const resolvers: Resolvers = {
   },
 
   Plan: {
+    // The user who owns/created the plan
+    planOwner: async (parent: Plan, _, context: MyContext): Promise<User> => {
```
briri: this should be labeled as `planCreator` and probably isn't what we want, especially when we add functionality to allow a plan to be "transferred", which is likely just going to be inviting a new collaborator.

The code always adds the plan creator as a `planCollaborator` with an `accessLevel` of `OWN`.

Check with Becky, but I think it would be ok to send an email to every collaborator with an `accessLevel === 'OWN'`

briri (changes requested):

src/resolvers/plan.ts
jupiter007: Ah good point. I'll check with Becky, but that makes sense.

jupiter007: @briri it looks like we specifically need the user that requested the feedback for the given `plan` so that the frontend can get the `affiliation.uri` for determining "readOnly" mode for feedback, unless the backend owns that logic.

However, Becky agreed that we will need to email all plan collaborators with `accessLevel` of `OWN` for the `completeFeedback` mutation

src/resolvers/feedback.ts
```diff
@@ -237,6 +240,21 @@ export const resolvers: Resolvers = {
...
             }
+
+            if (sendEmail) {
+              const planURL = `/projects/${project.id}/dmp/${planId}`;
+              const adminName = [context.token.givenName, context.token.surName]
+                .filter(Boolean).join(' ');
+
+              await sendFeedbackCompleteEmail(
+                context,
+                feedback.requestedById,
```
jupiter007, line 251: L251: @briri, as discussed, I updated `sendFeedbackCompleteEmail` to email the  `feedback requestor`.

src/resolvers/plan.ts
```diff
@@ -55,6 +56,10 @@ export const resolvers: Resolvers = {
         }
 
         const project = await Project.findById(reference, context, plan.projectId);
+        if (!project) {
```
jupiter007, line 59: Added this guard because `findById` can return `null`

src/resolvers/plan.ts
```diff
@@ -284,6 +289,13 @@ export const resolvers: Resolvers = {
   },
 
   Plan: {
+    // The user who owns/created the plan
+    planCreator: async (parent: Plan, _, context: MyContext): Promise<User> => {
```
jupiter007, line 293: L293: Changed this to `planCreator`

briri (approved):
Looks good @jupiter007 thanks for updating it

PR #715, opened 2026-05-05
Add ability to set the default Template

briri:
## Description

Part of [#213](https://github.com/CDLUC3/dmptool-doc/issues/213)

- Added an `isDefault` flag to Template and VersionedTemplate (table and models)
- Added a `markAsDefaultTemplate` mutation to allow SuperAdmins to mark a template as the default
- Added a `setDefaultTemplate` function to the `templateService`
- Added `isDefault` flag property to all query outputs

## Type of change

## How Has This Been Tested?

Added unit tests and updated tests for queries. Verified in the Apollo explorer

## Checklist:


src/services/templateService.ts
```diff
@@ -198,3 +199,70 @@ export const cloneTemplate = (
...
+  const tSQL = `UPDATE templates SET isDefault = ? WHERE id = ?;`;
+  const vtSQL = `UPDATE versionedTemplates SET isDefault = ? WHERE templateId = ?;`;
+
+  const currentDefault: VersionedTemplate = await VersionedTemplate.defaultTemplate(reference, context);
+
+  const newId = template.id.toString();
+  const oldId = currentDefault?.templateId?.toString();
+
+  // Mark the specified template as the default
+  context.logger.debug({ newDefault: template.id }, 'Setting new default template.');
```
jupiter007, line 227: L227: Just checking that you meant to leave this `console.log()` here and on L233

src/services/templateService.ts
```diff
@@ -198,3 +199,70 @@ export const cloneTemplate = (
...
+  if (tMarked) {
+    // Set the versionedTemplates
+    const vtMarked = await VersionedTemplate.query(context, vtSQL, ['1', newId], reference);
+
+console.log(`vtMarked: ${vtMarked}, oldId: ${oldId}`);
+
+    // If we did NOT successfully mark the new versioned templates
+    if (!vtMarked) {
+      // The marking of the versioned templates failed, so roll it back
+      context.logger.debug({newDefault: template.id}, 'Mark VersionedTemplate as default failed, rolling back changes.');
```
jupiter007, line 236: L236: I don't think this condition `!vtMarked` will ever happen because the query always returns an array.

src/services/templateService.ts
```diff
@@ -198,3 +199,70 @@ export const cloneTemplate = (
...
+    if (oldId) {
+      // Unmark the old template
+      context.logger.debug({newDefault: template.id}, 'Removing default designation from other templates.');
+      const tUnmarked = await Template.query(context, tSQL, ['0', oldId], reference);
+      // Unmark the old versionedTemplates
+      const vtUnmarked = await VersionedTemplate.query(context, vtSQL, ['0', oldId], reference);
+
+console.log(`vUnmarked: ${tUnmarked}, vtUnmarked: ${vtUnmarked}`);
+
+      if (!tUnmarked || !vtUnmarked) {
```
jupiter007: L253: Same here. I don't think `!tUnmarked` can ever be true since the `query` always returns an array. Is this check even necessary since you've already checked for `oldId`, which only exists if `currentDefault` exists?

src/resolvers/template.ts
```diff
@@ -229,6 +234,33 @@ export const resolvers: Resolvers = {
...
+    markAsDefaultTemplate: async (_, { templateId }, context: MyContext): Promise<Template> => {
+      const reference = 'markAsDefaultTemplate resolver';
+      try {
+        if (isSuperAdmin(context.token)) {
+          const template = await Template.findById(reference, context, templateId);
+          if (!template) {
+            throw NotFoundError('Template does not exist');
+          }
+
+          if (await setDefaultTemplate(reference, context, template)) {
```
jupiter007, line 247: L247: As mentioned in the templateService.ts file, `setDefaultTemplate` currently would not return `false` because the query always returns an array.

src/services/templateService.ts
```diff
@@ -198,3 +199,66 @@ export const cloneTemplate = (
...
+
+    // If there was a prior default
+    if (oldId) {
+      // Unmark the old template
+      context.logger.debug({newDefault: template.id}, 'Removing default designation from other templates.');
+      const tUnmarked = await Template.query(context, tSQL, ['0', oldId], reference);
+      // Unmark the old versionedTemplates
+      const vtUnmarked = await VersionedTemplate.query(context, vtSQL, ['0', oldId], reference);
+
+      if (!Array.isArray(tUnmarked) || tUnmarked.length < 1 || !Array.isArray(vtUnmarked) || vtUnmarked.length < 1) {
```
jupiter007: L249 and L234: I don't think we need `!Array.isArray(tUnmarked)` because the query always returns an array even when there is an error. I think we can just use `vtUnmarked.length === 0` because an empty array is returned when there is an error.

briri: Yep ok. I updated to just check the lengths

jupiter007 (approved):
Thanks for making the change @briri. Looks good.

PR #717, opened 2026-05-06
Updated requestFeedback resolver and emailService with new message

jupiter007:
## Description

- Updated `requestFeedback` resolver and `emailService` for that resolver with new message

Fixes # ([330](https://github.com/CDLUC3/dmptool-apollo-server/issues/330))

## Type of change

## How Has This Been Tested?
Manually tested and updated unit test

## Checklist:

### DIFF
--- a/src/resolvers/feedback.ts
+++ b/src/resolvers/feedback.ts
@@ -187,9 +187,17 @@ export const resolvers: Resolvers = {
             }
 
             const affiliation = await Affiliation.findByURI(reference, context, affiliationId);
-            // Send emails to the feedback recipients
-            await sendFeedbackRequestEmail(context, affiliation.feedbackEmails, messageToOrg ?? '');
 
+            if (affiliation.feedbackEmails.length === 0) {
+              context.logger.warn(prepareObjectForLogs({ affiliationId }), `Affiliation with ID ${affiliationId} has no feedback emails configured, so no notifications will be sent when requesting feedback`);
+            }
+
+            const planURL = `/projects/${project.id}/dmp/${planId}`;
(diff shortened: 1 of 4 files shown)


src/resolvers/feedback.ts
```diff
@@ -187,9 +187,17 @@ export const resolvers: Resolvers = {
             }
 
             const affiliation = await Affiliation.findByURI(reference, context, affiliationId);
-            // Send emails to the feedback recipients
-            await sendFeedbackRequestEmail(context, affiliation.feedbackEmails, messageToOrg ?? '');
 
+            if (affiliation.feedbackEmails.length === 0) {
```
briri, line 195: I forget, when the admin enables feedback for their org, are they required to provide feedback emails on the form or is it optional? 
 
If it's optional, lets fall back to the `affiliation.contactEmail` in this scenario. The contact email is required, so we should have something there. If they are required to provide feedback email(s) on that form, then this will be fine as is.

briri (approved):
Just had a clarifying question on whether or not the admin is required to fill out feedback emails on that page where they enable feedback and provide the message that gets displayed to users.

If they're not, and no feedback emails were provided, I think we should fall back to the `affiliation.contact_email`

If they're required to provide feedback email(s), then this is fine as is

src/resolvers/feedback.ts
jupiter007: Good question. Feedback email addresses are always required, along with the message:
image

briri: Ah ok. this looks goo to go then

PR #721, opened 2026-05-08
Seed data for RO Template w/plan

briri:
Add seed data that creates a published template with a Research Output question type and a Project+plan that is based off of the template.


jupiter007 (approved):
LGTM

PR #722, opened 2026-05-08
Added dedupe for re3 URIs in findRe3DataByURIs

jupiter007:
## Description
While working on a frontend bug fix for Research Outputs, I noticed that the `findRe3DataByURIs` was returning duplicate URIs, and restricting the number of repositories returned to just the number of different preferred URIs, so we might have two different repositories selected for "preferred", but the data returned would be duplicates of just the first URI.
- Added deduping to `findRe3DataByURIs` so that the frontend can get all of the expected `preferred repositories` in the search modal

Fixes # ([33](https://github.com/CDLUC3/dmptool-doc/issues/33))

## Type of change

## How Has This Been Tested?
Manually tested and updated some unit tests

## Checklist:


src/services/openSearchService.ts
```diff
@@ -284,7 +284,8 @@ export class OpenSearchService {
       response = await this.searchClient.search({
         index: 're3data',
         body: {
-          size: uris.length,
+          // Request more than uris.length to handle duplicate documents per URI
+          size: uris.length * 10,
```
briri: this seems fragile.

would be better to refactor this to perform pagination to get the exact number of records.

Something like: 
```
let allResults = [];
let cursor = null;
const pageSize = 100;

while (true) {
  const body = {
    size: pageSize,
    query: {
      terms: { uri: uris }
    },
    sort: [
      { "name.keyword": "asc" },
      { "_id": "asc" }  // Added to resolve dups
    ]
  };

  // Add cursor if we have one
  if (cursor) {
    body.search_after = cursor;
  }

  const response = await this.searchClient.search({
    index: 're3data',
    body: body
  });

  const hits = response.body.hits.hits;

  if (hits.length === 0) break; 

  allResults.push(...hits);

  // The very last hit is our cursor start
  cursor = hits[hits.length - 1].sort;
}
```

briri (changes requested):

src/services/openSearchService.ts
jupiter007: Thanks @briri. That totally makes sense. I've updated the `findRe3DataByURIs` per your suggestions.

briri: Thanks for making those changes @jupiter007. It looks good. Can you just remove the console.log debug statements?

briri (approved):
Approving. You can merge once the debug stuff has been cleaned up

PR #724, opened 2026-05-11
Add isDefault flag to MemberRole

briri:
## Description

Part of work for [#213](https://github.com/CDLUC3/dmptool-doc/issues/213)

- Add `isDefault` flag to `MemberRole` schema
- Add `setDefaultMemberRole` mutation to `MemberRole` schema and resolver. We had the ability for super admins to add, update and create Member Roles but not a specific way to set the default.

## Type of change

## How Has This Been Tested?

All tests pass. Verified I could change the default member role in apollo explorer. Verified that I could request the `isDefault` property in queries

## Checklist:

### DIFF
--- a/src/resolvers/memberRole.ts
+++ b/src/resolvers/memberRole.ts
@@ -5,11 +5,13 @@ import { MemberRole } from "../models/MemberRole";
 import { MyContext } from '../context';
 import { isSuperAdmin } from '../services/authService';
 import {
-  AuthenticationError, ForbiddenError, InternalServerError,
+  AuthenticationError,
+  ForbiddenError,
+  InternalServerError,
   NotFoundError
 } from '../utils/graphQLErrors';
 import { GraphQLError } from 'graphql';
-import {isNullOrUndefined, normaliseDateTime} from "../utils/helpers";
+import { isNullOrUndefined, normaliseDateTime } from "../utils/helpers";
 
 export const resolvers: Resolvers = {
   Query: {
--- a/src/resolvers/memberRole.ts
+++ b/src/resolvers/memberRole.ts
@@ -4,9 +4,12 @@ import { Resolvers } from "../types";
 import { MemberRole } from "../models/MemberRole";
 import { MyContext } from '../context';
 import { isSuperAdmin } from '../services/authService';
-import { AuthenticationError, ForbiddenError, InternalServerError } from '../utils/graphQLErrors';
+import {
+  AuthenticationError, ForbiddenError, InternalServerError,
+  NotFoundError
+} from '../utils/graphQLErrors';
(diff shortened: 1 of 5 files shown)


src/resolvers/memberRole.ts
```diff
@@ -113,6 +118,46 @@ export const resolvers: Resolvers = {
         throw InternalServerError();
       }
     },
+
+    // Set the default member role
+    setDefaultMemberRole: async (_, { id }, context) => {
+      const reference = 'setDefaultMemberRole resolver';
+      const original = await MemberRole.findById(reference, context, id);
```
jupiter007, line 125: L125: Do we need this initial `original` variable set, since we're setting it on L136?

Also, we should move the `isNullOrUndefined` from L126 after L136 then.

src/resolvers/memberRole.ts
```diff
@@ -113,6 +118,46 @@ export const resolvers: Resolvers = {
...
+          const newOne = await MemberRole.findById(reference, context, id);
+
+          try {
+            // Unset the original and then set the new one
+            await context.dataSources.sqlDataSource.query(context, unsetSql, []);
+            await context.dataSources.sqlDataSource.query(context, setSql, [id.toString()]);
+            newOne.isDefault = true;
+          } catch (err) {
+            context.logger.error(err, "Unable to set the default member role");
+            // Something went wrong so revert the changes
```
jupiter007, line 146: There might be an issue with the rollback.

If `unsetSql` failed, then `setSql` was never called and the database didn't update. But we still run `unsetSql` again in the `catch`, which could fail again.

I was reading about `transaction wrappers` and we could use something like: 
```
  public async transaction<T>(
    context: MyContext,
    fn: (conn: mysql2.PoolConnection) => Promise<T>
  ): Promise<T> {
    await this.initPromise;
    const connection = await this.getConnection();
    try {
      await connection.beginTransaction();
      const result = await fn(connection);
      await connection.commit();
      return result;
    } catch (err) {
      await connection.rollback();
      context.logger.error(prepareObjectForLogs(err), 'Transaction rolled back');
      throw err;
    } finally {
      connection.release();
    }
  }

```
 in `src/datasources/mysql.ts` and then do something like 
```
          try {
            await context.dataSources.sqlDataSource.transaction(context, async (conn) => {
              await conn.execute(unsetSql, []);
              await conn.execute(setSql, [id.toString()]);
            });
```

jupiter007: Or we could create separate `try-catch` for `unsetSql` and `setSql`

briri: good find. I was thinking about seeing if we could refactor things the other day to make use of SQL transactions (would be useful in a lot of places).

PR #731, opened 2026-05-13
Fixed some missing values for templates and versionedTemplates tables

jupiter007:
## Description

- Fixed error in `data-migrations/local-only/2026-05-08-1111-seed-project-plan.sql` when inserting data for `templates` and `versionedTemplates` tables which were missing the new `isDefault` value.


briri (approved):
thanks @jupiter007

PR #732, opened 2026-05-14
Updates to support REST API

briri:
## Description

Discovered that the DMP id and other identifiers being generated for the maDMP JSON records were incorrect, this update fixes those issues.
- Add additional `dmpId` argument to the `saveMaDMPVersion` function - The `dmp_id` on the record uses the UI endpoint (e.g. `/projects/123/dmp/123`) before the plan is published and the `dmpId` afterward. The code was doing that properly, but using the value for the DynamoDB Partition Key (PK) as well which we do not want. 
- Also updated it to pass the application name a `dmptool` or `dmptool-dev` instead of `DMP Tool (dev)`

Also needed the following changes:
- Add override for the protobufjs package 
- Add additional fields to the responses from the Affiliation queries
- Added `isDefault` flag to the MemberRole schema

## Type of change

## How Has This Been Tested?

Through the new REST API and Apollo explorer

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 - Added new `setDefaultMemberRole` mutation to member roles resolver.
+- Added override for protobufjs
 - Added `findByDMPId` to plan resolver
 - Added `defaultTemplate` query to the `versionedTemplate` schema and a corresponding resolver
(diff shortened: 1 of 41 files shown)


jupiter007 (approved):
LGTM

PR #737, opened 2026-05-19
Update moderate security vulnerabilities

briri:
Updated some moderate security vulnerabilities

- Update `jest` which is using an outdated version of `brace-expanision`
- Update `@graphql-codegen` which was using an outdated version of `ws`
- Updated our `@dmptool` dependencies
- Add overrides for `brace-expansion` and `ws`
- Removed overrides that were no longer needed: `fast-xml-parser`, `protobufjs` and `@node-oauth/oauth2-server`
- Update the `codegen.ts` config to coerce URL scalar types into strings
- Update the `RelatedWork` resolver to ensure OpenSearch hash arguments are strings. Typescript was flagging them

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v1.1.0
 
 ### Added
+- Added overrides for brace-expansion and ws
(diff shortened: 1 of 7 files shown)


jupiter007 (approved):
LGTM

PR #746, opened 2026-05-29
Updates for project search

jupiter007:
## Description

- Updated `dmphubAPI.ts` to check whether token is expired before each API call and updated `express.ts` to include `dmphubAPIDataSource` in the buildContext so that its available for the `searchExternalProjects` query
- Updated `searchExternalProjects` query to include structured `members` data in the response and the award year extracted from the project startDate
- Updated `projectImport` mutation to check if the funding record already exists. If so, then update it instead of creating a new one to avoid duplicates
- Updated `ExternalMember` schema to include `role`
Fixes # ([352](https://github.com/CDLUC3/dmptool-ui/issues/352))

## Type of change

## How Has This Been Tested?
Manually tested with frontend changes

## Checklist:

### DIFF
--- a/src/datasources/dmphubAPI.ts
+++ b/src/datasources/dmphubAPI.ts
@@ -65,9 +65,10 @@ export class Authorizer extends RESTDataSource {
     return Authorizer.#instance;
   }
 
-  // Initialize the Authorizer singleton
+  // Initialize the Authorizer singleton. Check whether token is expired before each API call
+  // and reauthenticate if necessary. This ensures we always have a valid token without needing to
   async init() {
-    if (!this.initialized) {
+    if (!this.initialized || this.hasExpired()) {
       await this.authenticate();
       this.initialized = true;
     }
@@ -90,12 +91,11 @@ export class Authorizer extends RESTDataSource {
 
   // Attach all of the necessary HTTP headers and the body prior to calling the token endpoint
   override willSendRequest(_path: string, request: AugmentedRequest) {
-    request.headers['authorization'] =`Basic ${this.creds}`;
+    request.headers['authorization'] = `Basic ${this.creds}`;
     request.headers['content-type'] = 'application/x-www-form-urlencoded';
     request.body =
       `grant_type=client_credentials&scope=` +
-      `${this.baseURL}/${this.env}.read` +
-      `${this.baseURL}/${this.env}.write`;
+      `${this.baseURL}/${this.env}.read ${this.baseURL}/${this.env}.write`;
   }
 }
 
@@ -151,23 +151,23 @@ export class DMPHubAPI extends RESTDataSource {
 
       // Build query parameters
       const params = new URLSearchParams();
-      if(!isNullOrUndefined(awardId)) {
+      if (!isNullOrUndefined(awardId)) {
         params.set("project", awardId);
       }
-      if(!isNullOrUndefined(piNames) && piNames.length) {
+      if (!isNullOrUndefined(piNames) && piNames.length) {
         params.set("pi_names", piNames.join(","));
       }
-      if(!isNullOrUndefined(awardName) && awardName.length) {
+      if (!isNullOrUndefined(awardName) && awardName.length) {
         params.set("keywords", awardName);
       }
-      if(!isNullOrUndefined(awardYear)) {
+      if (!isNullOrUndefined(awardYear)) {
         params.set("years", awardYear);
       }
 
       // Create path
       let path = apiTarget;
       const queryString = params.toString();
-      if(queryString){
+      if (queryString) {
         path += `?${queryString}`;
       }
 
@@ -184,9 +184,9 @@ export class DMPHubAPI extends RESTDataSource {
         `${reference} Error retrieving Awards from DMPHub API`
       );
       return null;
-    } catch(err) {
+    } catch (err) {
       context.logger.error(prepareObjectForLogs(err), `${reference} error calling DMPHub API getAwards`);
-      throw(err);
+      throw (err);
     }
   }
 }
--- a/src/middleware/express.ts
+++ b/src/middleware/express.ts
@@ -12,7 +12,7 @@ export async function attachApolloServer(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   cache: any,
   logger: Logger,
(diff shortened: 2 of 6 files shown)


src/datasources/dmphubAPI.ts
```diff
@@ -65,9 +65,10 @@ export class Authorizer extends RESTDataSource {
     return Authorizer.#instance;
   }
 
-  // Initialize the Authorizer singleton
+  // Initialize the Authorizer singleton. Check whether token is expired before each API call
+  // and reauthenticate if necessary. This ensures we always have a valid token without needing to
```
jupiter007, line 69: I noticed when testing that suddenly I was getting errors saying I was not authorized even though my client had an auth token. It was because of this initialization

src/middleware/express.ts
```diff
@@ -34,6 +34,7 @@ export async function attachApolloServer(
         cache,
         req.auth as JWTAccessToken,
         sqlDataSource,
+        dmphubAPIDataSource
```
jupiter007, line 37: The `searchExternalProjects` was setting dmphumAPi as null because `dmphumAPIDataSource` was not included in the context.

src/resolvers/project.ts
```diff
@@ -149,13 +149,28 @@ export const resolvers: Resolvers = {
             awardYear,
             piNames,
           );
+
           return dmps.map((dmpHubAward) => {
-            const members = [];
+            const members = Array.isArray(dmpHubAward.contributor)
```
jupiter007, line 155: This query was previously returning the default `members=[]` empty array, so I populated it for the response

src/resolvers/project.ts
```diff
@@ -149,13 +149,28 @@ export const resolvers: Resolvers = {
...
+                  role: contrib.role || []
+                };
+              })
+              : [];
 
             return {
               title: dmpHubAward.project.title,
               abstractText: dmpHubAward.project.description,
               startDate: normaliseDate(dmpHubAward.project.start),
+              awardYear: dmpHubAward.project.start ? dmpHubAward.project.start.slice(0, 4) : null,
```
jupiter007, line 180: For now I am returning the `awardYear` extracted from the `startDate` until we figure out how to get the actual `award` year?

src/resolvers/project.ts
```diff
@@ -310,9 +325,30 @@ export const resolvers: Resolvers = {
               const addFundingErrors = [];
               for (const fund of input.funding) {
                 const newFunding = new ProjectFunding(fund);
-                const fundingAdded = await newFunding.create(context, projectId);
-                if (!fundingAdded) {
-                  addFundingErrors.push(`Funding(affiliationId=${newFunding.affiliationId})`);
+
+                // Check if a funding record already exists for this affiliation on the project. If so, 
+                // update it instead of creating a new one to avoid duplicates.
+                const existingFunding = await ProjectFunding.findByProjectAndAffiliation(
```
jupiter007, line 338: `projectImport` was always inserting a new funding record instead of checking whether we should just be updating it.

src/schemas/project.ts
```diff
@@ -244,6 +244,8 @@ export const projectTypeDefs = gql`
     orcid: String
     "The member's email address"
     email: String
+    "The member's role(s) in the project (e.g. PI, Co-PI, Research Assistant, etc.)"
+    role: [String]
```
jupiter007, line 248: Added `role` to the `members` even though Jamie assigned all contributors as `investigation`, which we can address in a later ticket

src/resolvers/project.ts
```diff
@@ -149,13 +149,28 @@ export const resolvers: Resolvers = {
             awardYear,
             piNames,
           );
+
           return dmps.map((dmpHubAward) => {
-            const members = [];
+            const members = Array.isArray(dmpHubAward.contributor)
+              ? dmpHubAward.contributor.map(contrib => {
+                const [email, ...nameParts] = contrib.name.split(',');
```
briri: I'm not sure this is what you want. The contributor / contact are specific types [defined here](https://github.com/CDLUC3/dmptool-utils/blob/b399c2367c846443b630adb11720342907e95767/src/maDMPTypes.ts#L118)

If you're seeing the person's email in the `name` part, it won't necessarily always be there because we have 3 different APIs returning data. 
Here are how those names are generated in each scenario: 
- [Crossref API](https://github.com/CDLUC3/dmsp_api_prototype/blob/0913fb72464e28749e4a43437814b730491f1df3/lambdas/api/get_awards_crossref/app.rb#L145)
- [NIH API](https://github.com/CDLUC3/dmsp_api_prototype/blob/0913fb72464e28749e4a43437814b730491f1df3/lambdas/api/get_awards_nih/app.rb#L125)
- [NSF API](https://github.com/CDLUC3/dmsp_api_prototype/blob/0913fb72464e28749e4a43437814b730491f1df3/lambdas/api/get_awards_nsf/app.rb#L172)

At the very least you might want to add something that verifies that the value looks like an email.

The DMPHub Award has a contact which possibly contains an `mbox` which would be a verified email

src/resolvers/project.ts
briri: 👍🏻 yeah, I'm not sure either. Crossref has an actual `award-start`, NSF has a `startDate` and NIH has a `project_start_date`.

src/schemas/project.ts
briri: PIs are all we get (or used to get) from the funder APIs, so I think it may be all we ever get

briri (changes requested):
I think we need to handle the names/email a bit differently. I added some comments and links to the Lambdas that make the API calls to the funders.

src/resolvers/project.ts
jupiter007: @briri I consistently see "contributor" object being returned like:
```
{
apollo-server  |   name: 'mrepasch@unm.edu, Repasch, N, Marisa',
apollo-server  |   role: [ 'http://credit.niso.org/contributor-roles/investigation' ]
apollo-server  | }
```

That's why I used that split. 

But I added validation to make sure we are getting what we expect.

jupiter007: Also, I checked the `mbox` and it is partially masked like `h****s@ucsd.edu`

briri (approved):
Thanks for adding that email check

PR #749, opened 2026-06-01
Update dependencies

briri:
- Updated `uuid` and `@testcontainers/mysql`
- Added override for `qs`
- Removed overrides for `ws` and `brace-expansion` as they were no longer necessary


jupiter007 (approved):
LGTM. Thanks Brian.

PR #752, opened 2026-06-01
Updated PlanSectionProgress class to return totalRequiredQuestions and answeredRequiredQuestions 

jupiter007:
## Description

- Updated methods for the `PlanSectionProgress` class to return `totalRequiredQuestions` and `answeredRequiredQuestions`, and updated unit tests and `PlanSectionProgress` schema

Fixes # ([249](https://github.com/CDLUC3/dmptool-apollo-server/issues/249))

## Type of change

## How Has This Been Tested?
Manually tested and updated unit tests

## Checklist:

### DIFF
--- a/src/models/Plan.ts
+++ b/src/models/Plan.ts
@@ -135,6 +135,8 @@ export class PlanSectionProgress {
   public displayOrder: number;
   public totalQuestions: number;
   public answeredQuestions: number;
+  public totalRequiredQuestions: number;
+  public answeredRequiredQuestions: number;
   public tags?: Tag[];
 
   constructor(options) {
@@ -145,6 +147,8 @@ export class PlanSectionProgress {
     this.displayOrder = options.displayOrder;
     this.totalQuestions = options.totalQuestions;
     this.answeredQuestions = options.answeredQuestions;
+    this.totalRequiredQuestions = options.totalRequiredQuestions ?? 0;
+    this.answeredRequiredQuestions = options.answeredRequiredQuestions ?? 0;
     this.tags = options.tags ?? [];
   }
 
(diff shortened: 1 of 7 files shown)


src/models/Plan.ts
```diff
@@ -295,6 +302,11 @@ export class PlanSectionProgress {
           WHEN a.id IS NOT NULL AND JSON_TYPE(a.json) = 'OBJECT'
           THEN vq.id
         END) AS answeredQuestions,
+      COUNT(DISTINCT CASE WHEN vq.required = 1 THEN vq.id END) AS totalRequiredQuestions,
+      COUNT(DISTINCT CASE
```
briri, line 317: Now that I'm looking at both this new field and the old `answeredQuestions` one. I'm wondering if this will give false positives. If for example the answer is a `textArea` type and has the correct JSON object structure but the actual value is a blank string. We probably do not want to count that as being "answered".

I think you can get that scenario if you first answer a question (to create the answer record), and then go back and delete the text.

See if you can add something like `AND TRIM(JSON_UNQUOTE(JSON_EXTRACT(a.json, '$.answer'))) != ''` I think you can use `JSON_ARRAY()` and `JSON_OBJECT()` as well to test if `answer` is an empty array or object.

briri (commented):
This looks good @jupiter007.

I flagged a potential issue though that I think can occur (not just with your new required field but the existing `answeredQuestions` as well).

src/models/Plan.ts
jupiter007: Great catch @briri. I needed to add special handling for `text` and `textArea` questions.

I've made those updates, and they appear to be working

briri (approved):
Looks good. Thanks for making that change

Issue #461, opened 2025-09-30
Create a resolver to handle Organization Details form submittal, as well as handle org logos


jupiter007:
A static Organization Details page was created (view details here: https://github.com/CDLUC3/dmsp_frontend_prototype/issues/11), and the form includes a Logo upload.

We need a mutation resolver to handle all the info submitted in the form.

I read that uploading binary files through the GraphQL mutation itself bloats requests and forces the server to buffer and forward large file payloads, which is inefficient and can slow things down.

Would it be possible to use a `presigned url` for logos?

PR #753, opened 2026-06-02
Add ability to upload logos

briri:
## Description

Part of [#461](https://github.com/CDLUC3/dmptool-apollo-server/issues/461)

Adds the ability to upload affiliation logos.

An upload involves 3 steps on the part of the caller:
1. Call `generateLogoUploadURL` to get a presigned URL for S3
2. Upload the file to S3 directly
3. Call `finalizeLogoUpload` (assuming step 2 succeeded) to update the Affiliation record with the logo's S3 key (found in the response from step 1)

- Added the `generateLogoUploadURL` and `finalizeLogoUpload` resolvers to the `affiliation` schema.
- Added new data migration to remove `logoURI` column from `affiliations`. The URI is now generated on the fly by the resolver.
- Added new `src/datasource/s3.ts` file to provide the CDN URL and to generate presigned URLs for S3 objects.
- Added `s3` properties to the `awsConfig` object
- Added `src/datasources/s3.ts` to handle generation of presigned URLs
- Updated `awsConfig` to move SES properties underneath the `ses` property

## Type of change

## How Has This Been Tested?

Via unit tests and the Apollo explorer

Full end-to-end testing will be done when we start on the UI portion of this code. I expect we may need to make some small adjustments.

## Checklist:


data-migrations/2026-06-01-1542-remove-logoURI-from-affiliations.sql
```diff
@@ -0,0 +1,3 @@
+-- Remove the logoURI column from the affiliation table, as it is now constructed
+-- from the CDN URL and logoName (s3 key) within the resolver.
+ALTER TABLE `affiliations` DROP COLUMN `logoURI`;
```
briri, line 3: Found that we can just do this via a chained resolver. This will be better since the URL or S3 bucket can change and we won't need to run a DB migration to update the data.

localstack-setup/init-aws.sh
```diff
@@ -95,6 +97,9 @@ if [ -e "$GENERATE_MADMP_RECORD_FUNCTION" ]; then
     --batch-size 1 \
     --event-source-arn $QUEUE_ARN
 
+  # Create an S3 bucket that we will use to store logos and other media assets for the DMPTool
+  awslocal s3 mb s3://$S3_BUCKET_NAME
```
briri, line 101: Creates an S3 bucket where logos can be stored locally. Note that they will likely not persist after the docker stack is shut down.

src/config/awsConfig.ts
```diff
@@ -27,12 +27,19 @@ export const awsConfig = {
   region: process.env.AWS_REGION || 'us-west-2',
 
   // Simple Email Service (SES) configuration
-  sesEndpoint: process.env.SES_ENDPOINT,
-  port: process.env.SES_PORT ?? 465,
-  sesAccessKey: process.env.SES_ACCESS_ID,
-  sesAccessSecret: process.env.SES_ACCESS_SECRET,
-  sesBounceAddress: process.env.SES_BOUNCE_EMAIL_ADDRESS,
-  sesBouncedEmailBucket: process.env.SES_BOUNCED_EMAIL_BUCKET,
+  ses: {
```
briri, line 30: I liked the nested pattern Jamie introduced better.

src/datasources/s3.ts
```diff
@@ -0,0 +1,81 @@
...
+  affiliationURI: string,
+  fileName: string,
+  contentType = 'image/png'
+): Promise<{ url: string, fields: string }> => {
+  const bucket = awsConfig.s3.bucket;
+
+  try {
+    const uri = URL.parse(affiliationURI);
+    const path = uri.pathname.startsWith('/') ? uri.pathname.slice(1) : uri.pathname;
+    const logoKey = `logos/${uri.host}/${path}/${Date.now()}-${fileName}`;
```
briri: setting the S3 key to include the affiliation id. I think this will make things easier if we ever need to go look at things manually in the bucket

src/resolvers/affiliation.ts
```diff
@@ -196,9 +220,78 @@ export const resolvers: Resolvers = {
...
+          return affiliation;
+        }
+
+        return updated;
+      }),
   },
 
   Affiliation: {
+    logoURI: (parent: Affiliation): string => {
+      return `${CDN_BASE_URL}${parent.logoName}`;
```
briri, line 301: Simple resolver to append the S3 key, `logoName` to the CDN URL

src/services/authService.ts
```diff
@@ -27,7 +27,8 @@ export const authenticatedResolver = <TResult, TArgs, TParent = Record<PropertyK
 ): AuthenticatedResolverType<TResult, TArgs, TParent> => {
   return async (parent: TParent, args: TArgs, context: MyContext, info: GraphQLResolveInfo): Promise<TResult> => {
     try {
-      if ((authLevel === UserRole.ADMIN && !isAdmin(context.token))
+      if ((isNullOrUndefined(context.token)
```
briri, line 30: Noticed that we could add an initial check to verify that there is a token at all before processing the other role checks

.gitignore
```diff
@@ -1,6 +1,9 @@
 # Ignore the Dotenv file
 .env
 
+# Ignore any local packages
+*.tgz
```
briri, line 5: Used `npm pack` to tarball the `@dmptool/utils` package for testing, and then copied it into the project root of this repo. 

Doing it that way makes it available when running inside the docker image.

docker-compose.yaml
```diff
@@ -143,8 +143,8 @@ services:
       - "DYNAMODB_TABLE_NAME=localDMPTable"
       - "MAX_ATTEMPTS=3"
 
-      # LocalStack managed SQS queue
```
briri, line 146: We switched to update DynamoDB directly, so removing this unused ENV file

Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN pip install --no-cache-dir awscli-local
 
 # Copy package.json and package-lock.json
 # to the /app working directory
-COPY package*.json tsconfig.json codegen.ts .env ./
+COPY package*.json tsconfig.json codegen.ts .env dmptool-*.tgz ./
```
briri, line 36: This will allow us to use `npm pack` to test out changes to `@dmptool/types` and `@dmptool/utils` more easily

data-migrations/2026-06-01-1542-remove-logoURI-from-affiliations.sql
jupiter007: If you run this before the `local` data-migration scripts, wouldn't the inserts into the `affiliations` table fail when trying to insert a value for `logoURI`?

src/datasources/s3.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Step 3: The caller sends a mutation request to finalize the upload. This allows
+ *         the caller to handle any errors that have occurred before we store
+ *         the location of the newly uploaded object in the database.
+ */
+
+/**
+ * The URL to the application's CDN (AWS Cloudfront distribution)
+ */
+export const CDN_BASE_URL: string = process.env.NODE_ENV === 'development'
+  ? `https://${awsConfig.s3.bucket}.s3.us-west-2.amazonaws.com`
```
jupiter007: L42: Shoud `us-web-2.amazonaws.com` come from `awsConfig` as well instead of hard-coding it here?

src/datasources/s3.ts
```diff
@@ -0,0 +1,81 @@
...
+export const getPresignedURLForAffiliationLogo = async(
+  logger: Logger,
+  affiliationURI: string,
+  fileName: string,
+  contentType = 'image/png'
+): Promise<{ url: string, fields: string }> => {
+  const bucket = awsConfig.s3.bucket;
+
+  try {
+    const uri = URL.parse(affiliationURI);
```
jupiter007, line 106: L64: Do you need a null check for `uri`because the `uri` could be malformed and return a null from parsing.

src/datasources/s3.ts
jupiter007: L66: Are you sure you want to use `${Date.now()}` in the key? How will you be able to clean it up later?

Also, do we need to sanitize the `fileName`?

src/resolvers/affiliation.ts
```diff
@@ -196,9 +220,78 @@ export const resolvers: Resolvers = {
...
+     * @param _ Ignored, this is the entrypoint for the Apollo resolver
+     * @param args the affiliation URI and the name of the logo (S3 key)
+     * @param context The Apollo context
+     * @returns The Affiliation (with errors if applicable)
+     * @throws NotFoundError when the Affiliation is not found
+     * @throws ForbiddenError when the caller does not have permission
+     * @throws UnauthorizedError when the JWT token is not present
+     * @throws InternalServerError when a fatal error occurred
+     */
+    finalizeLogoUpload: authenticatedResolver(
```
jupiter007, line 269: `authenticatedResolver` doesn't check whether the user is an Admin for the given `affiliation URI`.

data-migrations/local-only/2025-11-20-0349-seed-affiliations.sql
```diff
@@ -7,7 +7,7 @@ SET @default_super_id := (SELECT id FROM userEmails WHERE email = 'super@example
 SET @default_admin_id := (SELECT id FROM userEmails WHERE email = 'admin@example.com');
 
 -- Default CDL Affiliation with departments
-INSERT INTO `affiliations` (`uri`, `provenance`, `name`, `displayName`, `searchName`, `funder`, `fundrefId`, `homepage`, `acronyms`, `aliases`, `types`, `logoURI`, `logoName`, `contactName`, `contactEmail`, `ssoEntityId`, `feedbackEnabled`, `feedbackMessage`, `feedbackEmails`, `managed`, `createdById`, `created`, `modifiedById`, `modified`) VALUES ('https://ror.org/03yrm5c26', 'ROR', 'California Digital Library', 'California Digital Library (cdlib.org)', 'California Digital Library | cdlib.org | CDL ', 0, NULL, 'http://www.cdlib.org/', '["CDL"]', '[]', '["Archive"]', NULL, NULL, 'UC3 Helpdesk', 'uc3@cdlib.org', NULL, false, '<p>Dear %{user_name},</p><p>"%{plan_name}" has been sent to your %{application_name} account administrator for feedback.</p><p>Please email %{organisation_email} with any questions about this process.</p>', '["uc3@cdlib.org"]', true, @default_super_id, CURDATE(), @default_super_id, CURDATE());
```
briri, line 10: Had to remove the `logoURI` from here too

docker-compose.yaml
```diff
@@ -76,6 +76,7 @@ services:
       - "127.0.0.1:4510-4559:4510-4559"  # external services port range
     environment:
       # LocalStack configuration: https://docs.localstack.cloud/references/configuration/
+      - LOCALSTACK_AUTH_TOKEN=${LOCALSTACK_AUTH_TOKEN- }
```
briri, line 79: I found that I needed to do this to get the docker stack running

src/resolvers/affiliation.ts
```diff
@@ -196,9 +220,78 @@ export const resolvers: Resolvers = {
...
+      async (
+        _: Record<PropertyKey, never>,
+        { affiliationURI, fileName, contentType }: {
+          affiliationURI: string,
+          fileName: string,
+          contentType: string
+        },
+        context: MyContext
+      ): Promise<AffiliationLogoUpload> => {
+        return await getPresignedURLForAffiliationLogo(context.logger, affiliationURI, fileName, contentType);
```
jupiter007: Do we need to limit the contentType here or return an error if it's not within the types we accept?

src/datasources/s3.ts
```diff
@@ -0,0 +1,81 @@
...
+      bucket,
+      logoKey,
+      contentType,
+      awsConfig.region || 'us-west-2'
+    );
+    logger.debug({ affiliationURI, fileName, bucket, url, fields }, 'Generated presigned URL');
+
+    return { url, fields };
+  } catch (err) {
+    logger.error({ affiliationURI, fileName, bucket, err }, 'Unable to generate presigned URL for S3.');
```
jupiter007, line 124: Do we want to throw an error if this fails?

src/resolvers/affiliation.ts
jupiter007: L249: Could `getPresignedURLForAffiliationLogo return `null` or `undefined` and is that expected?

data-migrations/2026-06-01-1542-remove-logoURI-from-affiliations.sql
briri: yep. I updated that seed file for the affiliations table

src/datasources/s3.ts
briri: yeah. I'll use the region defined in the `awsConfig`

src/datasources/s3.ts
briri: Yes, I did some refactoring and added a function to use String's `normalize` and also limit the length of the file name

src/datasources/s3.ts
briri: I updated it to return undefined here and if the URI doesn't parse

src/resolvers/affiliation.ts
briri: [The `@dmptool/utils` function](https://github.com/CDLUC3/dmptool-utils/blob/9f16f76a3b9414dd5768cc15a886fc995ccf78d2/src/s3.ts#L204) enforces that the contentType start with `image/` and that whatever contentType was specified in the call to the resolver MUST match what's actually sent to S3. It's very strict.

src/resolvers/affiliation.ts
briri: Great catch!

jupiter007 (approved):

PR #762, opened 2026-06-15
Added support for Admin Notifications

jupiter007:
## Description

- Added `adminNotifications` model, resolver and db table. 
- Added `findByIdWithTemplateName` method to `TemplateCustomization` model so that we can return the template name 
- Updated `requestFeedback`, `addTemplate` and `publishTemplateCustomization` resolvers to add a record to the `adminNotifications` table

Fixes # ([570](https://github.com/CDLUC3/dmptool-ui/issues/570))

## Type of change

## How Has This Been Tested?
Manually tested and with updated and new unit tests

## Checklist:
* Frontend changes will be in PR review shortly

### DIFF
--- a/data-migrations/2026-06-11-0907-create-adminNotifications.sql
+++ b/data-migrations/2026-06-11-0907-create-adminNotifications.sql
@@ -1,5 +1,6 @@
 CREATE TABLE `adminNotifications` (
   `id` int unsigned NOT NULL AUTO_INCREMENT,
+  `userId` int unsigned NOT NULL,
   `notificationType` varchar(50) NOT NULL,
   `affiliationId` varchar(255) NOT NULL,
   `metadata` json DEFAULT NULL,
@@ -11,6 +12,8 @@ CREATE TABLE `adminNotifications` (
   PRIMARY KEY (`id`),
   KEY `affiliationId` (`affiliationId`),
   KEY `notificationType` (`notificationType`),
+  KEY `userId` (`userId`),
+  CONSTRAINT `adminNotifications_ibfk_5` FOREIGN KEY (`userId`) REFERENCES `users` (`id`),
   CONSTRAINT `adminNotifications_ibfk_2` FOREIGN KEY (`affiliationId`) REFERENCES `affiliations` (`uri`),
   CONSTRAINT `adminNotifications_ibfk_3` FOREIGN KEY (`createdById`) REFERENCES `users` (`id`),
   CONSTRAINT `adminNotifications_ibfk_4` FOREIGN KEY (`modifiedById`) REFERENCES `users` (`id`)
--- a/data-migrations/2026-06-11-0907-create-adminNotifications.sql
+++ b/data-migrations/2026-06-11-0907-create-adminNotifications.sql
@@ -1,0 +1,17 @@
+CREATE TABLE `adminNotifications` (
+  `id` int unsigned NOT NULL AUTO_INCREMENT,
+  `notificationType` varchar(50) NOT NULL,
+  `affiliationId` varchar(255) NOT NULL,
+  `metadata` json DEFAULT NULL,
+  `isRead` tinyint(1) NOT NULL DEFAULT 0,
+  `createdById` int unsigned NOT NULL,
+  `created` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `modifiedById` int unsigned NOT NULL,
+  `modified` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  PRIMARY KEY (`id`),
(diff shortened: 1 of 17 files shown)


src/models/TemplateCustomization.ts
```diff
@@ -580,6 +580,8 @@ export class TemplateCustomization extends MySqlModel {
   public latestPublishedDate: string;
   // Whether the customization has been modified since it was last published
   public isDirty: boolean;
+  // The name of the parent template, included for convenience when fetching a customization with its template name
+  public templateName?: string;
```
jupiter007, line 584: I added `templateName` as optional since it would be nice to get the name of the template that was customized

src/resolvers/adminNotifications.ts
```diff
@@ -0,0 +1,219 @@
...
+          context,
+          affiliationId,
+          paginationOptions
+        );
+      }
+    ),
+  },
+
+  Mutation: {
+    addAdminNotification: authenticatedResolver(
```
jupiter007: Currently, this `addAdminNotification` is not used, since we only create adminNotification records from resolvers like `requestFeedback`. But I added it just in case we'll want to add notifications in other ways

src/resolvers/templateCustomization.ts
```diff
@@ -6,6 +6,7 @@ import {
 import { authenticatedResolver } from "../services/authService";
 import { MyContext } from "../context";
 import { VersionedTemplate } from "../models/VersionedTemplate";
+import { AdminNotification } from "../models/AdminNotifications";
```
jupiter007, line 9: Sorry for all these `prettier` formatting changes. The only change was to `publishTemplateCustomization`

src/models/AdminNotifications.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  static async findByAffiliationId(
+    reference: string,
+    context: MyContext,
+    affiliationId: string | null,
+    options: PaginationOptions = AdminNotificationResults.getDefaultPaginationOptions(),
+    isRead?: boolean,
+  ): Promise<PaginatedQueryResults<AdminNotificationResults>> {
+
+    // SuperAdmins get all notifications, Admins only get their affiliation's notifications
```
briri: Check with Becky, but I think Super Admins son't want to see all notifications. It could get very noisy.

data-migrations/2026-06-11-0907-create-adminNotifications.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE `adminNotifications` (
+  `id` int unsigned NOT NULL AUTO_INCREMENT,
+  `notificationType` varchar(50) NOT NULL,
+  `affiliationId` varchar(255) NOT NULL,
+  `metadata` json DEFAULT NULL,
+  `isRead` tinyint(1) NOT NULL DEFAULT 0,
```
briri, line 7: I think we want a `userId` here. Many orgs have multiple admins. If one marks a notification as "read" then it would apply for the others and we probably want them to be able to control their own notifications individually.

src/models/AdminNotifications.ts
```diff
@@ -0,0 +1,192 @@
...
+        if (updated) {
+          return await AdminNotification.findById(reference, context, this.id);
+        }
+      }
+    } else {
+      this.addError('general', 'AdminNotification has never been saved');
+    }
+    return this;
+  }
+
```
briri, line 181: We probably want a static `addNotificationForAffiliation` with a SQL query like 
```sql
INSERT INTO (userId, notificationType, a, b, c) 
(SELECT u.id, ?, ?, Date(), ? 
 FROM users AS u 
 WHERE u.role = 'ADMIN' AND u.affiliationId = ?);
```

briri (changes requested):
I think we need these to be per user not per affiliation

src/models/AdminNotifications.ts
jupiter007: @briri I confirmed with Becky that we should treat SuperAdmins like any other user, so I removed that scenario

data-migrations/2026-06-11-0907-create-adminNotifications.sql
jupiter007: Thanks for clarifying Brian. That totally makes sense. I added `userId` to the table and made updates to search on `userId`

src/models/AdminNotifications.ts
jupiter007: Thanks. I added that function and included SUPERADMIN in the condition as well, since they will be treated like other admins and see any notifications just for their organization.

briri (approved):
thanks. Looks good

PR #771, opened 2026-06-22
added a research output question to the local default template

briri:
Added a data migration to set the default template to the DCC template, and then add a new research output section and question to that template and republish it.

This is needed for the new REST API so I can test how it handles the translation of the DMP narrative and RDA common standard datasets in JSON format


jupiter007 (approved):
LGTM

PR #773, opened 2026-06-25
fix bug in data migration to add research output question to default …

briri:
I just realized I forgot to checkin a bug fix for the research output question that was added to the default template for our local DBs.

The new published version of the default template only had the new research output question, not all of the existing questions!


jupiter007 (approved):
LGTM

PR #774, opened 2026-06-26
Updates for the Admin User Profile page

jupiter007:
## Description

- Added `isArchived` field to the `users` table to help us filter those users out when returned in `users` response
- Added `findByProjectIdWithPagination` method to the `PlanSearchResult` model for the `plans` resolver 
- Added `userProjects` query to return all projects for a specified user, with search term and pagination
- Added `updateUserRole` mutation for admins to change a user's role, `updateUserInfo` mutation for super admins to update a specified user's profile, and an `archiveUser` mutation to archive a user
- Updated `plans` resolver to have pagination for a specified `userId` with optional `search term`, and added a chained resolver for `PlanSearchResult` for `templateOwnerAffiliationName`
- Updated `PlanSearchResult` model to include `createdById` and `templateOwnerAffiliationName` for the Admin Users page

Fixes # ([281](https://github.com/CDLUC3/dmptool-doc/issues/281))

## Type of change

## How Has This Been Tested?
Manually tested and tested with new unit tests, and frontend changes.

## Checklist:
*Note: There are corresponding frontend changes: https://github.com/CDLUC3/dmptool-ui/pull/1322

### DIFF
--- a/data-migrations/2026-06-25-0919-add-isArchived-field-to-users.sql
+++ b/data-migrations/2026-06-25-0919-add-isArchived-field-to-users.sql
@@ -1,0 +1,2 @@
+ALTER TABLE `users`
+ADD COLUMN `isArchived` tinyint(1) NOT NULL DEFAULT '0' AFTER `active`;
--- a/src/models/Plan.ts
+++ b/src/models/Plan.ts
@@ -139,12 +139,14 @@ export class PlanSearchResult {
   }
 
   /**
-   * Find high-level details about the plans for a project. This information is
-   * meant to supply an overview of the plans.
+   * Find projects for a specified userId, with pagination and optional search term filtering. 
+   * This method returns a paginated list of PlanSearchResult objects that match the search criteria.
    *
    * @param reference The caller's reference string for logging purposes'
    * @param context The Apollo context object
-   * @param projectId The ID of the project to return plans for
+   * @param userId The ID of the user to return projects for
+   * @param options Pagination options for the query
+   * @param term Optional search term to filter the results
    * @returns An array of PlanSearchResult objects
    */
   static async findByProjectIdWithPagination(
--- a/src/models/Plan.ts
+++ b/src/models/Plan.ts
@@ -139,7 +139,7 @@ export class PlanSearchResult {
   }
 
   /**
-   * Find projects for a specified userId, with pagination and optional search term filtering. 
+   * Find projects/plans for a specified userId, with pagination and optional search term filtering. 
    * This method returns a paginated list of PlanSearchResult objects that match the search criteria.
    *
    * @param reference The caller's reference string for logging purposes'
@@ -149,7 +149,7 @@ export class PlanSearchResult {
    * @param term Optional search term to filter the results
    * @returns An array of PlanSearchResult objects
    */
-  static async findByProjectIdWithPagination(
+  static async findByUserIdWithPagination(
     reference: string,
     context: MyContext,
     userId: number,
--- a/src/models/Plan.ts
+++ b/src/models/Plan.ts
@@ -13,6 +13,14 @@ import { PlanGuidance } from "./Guidance";
 import { VersionedTemplate } from "./VersionedTemplate";
 import { Project } from "./Project";
 import { Tag } from "./Tag";
+import {
+  PaginatedQueryResults,
+  PaginationOptions,
+  PaginationOptionsForCursors,
+  PaginationOptionsForOffsets,
+  PaginationType
+} from '../types/general';
+import { prepareObjectForLogs } from '../logger';
 
 export const DEFAULT_TEMPORARY_DMP_ID_PREFIX = 'temp-dmpId-';
 
@@ -52,6 +60,7 @@ export class PlanSearchResult {
   public id: number;
   public createdBy: string;
   public created: string;
+  public createdById: number;
   public modifiedBy: string;
   public modified: string;
   public title: string;
@@ -62,6 +71,7 @@ export class PlanSearchResult {
   public members: string;
   public templateTitle: string;
   public versionedTemplateId: number;
+  public templateOwnerAffiliationName: string;
 
   // The following fields will only be set when the plan is published!
   public dmpId: string;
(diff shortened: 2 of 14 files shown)


data-migrations/2026-06-25-0919-add-isArchived-field-to-users.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `users`
+ADD COLUMN `isArchived` tinyint(1) NOT NULL DEFAULT '0' AFTER `active`;
```
jupiter007, line 2: I added a new `isArchived` field, so that it is easier to filter out these records for `users` queries, and seemed more reliable then checking name "Delete Account"

src/models/Plan.ts
```diff
@@ -125,6 +137,97 @@ export class PlanSearchResult {
...
+   * This method returns a paginated list of PlanSearchResult objects that match the search criteria.
+   *
+   * @param reference The caller's reference string for logging purposes'
+   * @param context The Apollo context object
+   * @param userId The ID of the user to return projects for
+   * @param options Pagination options for the query
+   * @param term Optional search term to filter the results
+   * @returns An array of PlanSearchResult objects
+   */
+  static async findByProjectIdWithPagination(
```
briri: Should this be `findByUserIdWithPagination`?

src/models/Plan.ts
```diff
@@ -125,6 +137,97 @@ export class PlanSearchResult {
     const results = await Plan.query(context, sql, [projectId?.toString()], reference);
     return Array.isArray(results) ? results.map((entry) => new PlanSearchResult(entry)) : [];
   }
+
+  /**
+   * Find projects for a specified userId, with pagination and optional search term filtering.
```
briri: Should be `Find plans for a specified userId ...` or `Find projects/plans for a specified userId ...`

src/resolvers/plan.ts
```diff
@@ -525,6 +542,30 @@ export const resolvers: Resolvers = {
...
+
+      const affiliation = await Affiliation.findByURI(
+        'planSearchResult.templateOwnerAffiliationName resolver',
+        context,
+        versionedTemplate.ownerId
+      );
+      return affiliation?.displayName || null;
+    },
+    user: async (parent: PlanSearchResult, _, context: MyContext): Promise<User> => {
+      console.log('****parent', parent);
```
briri: you can delete this `console.log`

src/resolvers/plan.ts
```diff
@@ -525,6 +542,30 @@ export const resolvers: Resolvers = {
...
+      if (!versionedTemplate?.ownerId) return null;
+
+      const affiliation = await Affiliation.findByURI(
+        'planSearchResult.templateOwnerAffiliationName resolver',
+        context,
+        versionedTemplate.ownerId
+      );
+      return affiliation?.displayName || null;
+    },
+    user: async (parent: PlanSearchResult, _, context: MyContext): Promise<User> => {
```
briri: This should probably be `createdBy` instead of `user` since there is also a `modifiedBy` and potentially a `registeredBy` so `user` would be unclear

src/resolvers/project.ts
```diff
@@ -109,6 +110,48 @@ export const resolvers: Resolvers = {
         throw InternalServerError();
       }
     },
+    // Return all projects for a specified user (Admin only!) with pagination and optional search term filtering
+    userProjects: authenticatedResolver(
```
briri, line 114: Fine to leave this, but how is it being used differently from the "plans by userId" that you updated above?

src/resolvers/user.ts
```diff
@@ -133,6 +136,131 @@ export const resolvers: Resolvers = {
...
+          if (otherAffiliationName) {
+            const affiliation = await processOtherAffiliationName(context, otherAffiliationName);
+            if (affiliation.hasErrors()) {
+              const err = affiliation.errors?.general ?? 'Unable to save the affiliation at this time';
+              user.addError('otherAffiliationName', err);
+              return user;
+            }
+            user.affiliationId = affiliation.uri;
+          } else {
+            user.affiliationId = affiliationId;
```
briri, line 174: If the `affiliationId` changes, we should revert the user's role to `RESEARCHER`.

src/resolvers/user.ts
```diff
@@ -417,6 +545,30 @@ export const resolvers: Resolvers = {
       }
     },
 
+    // Anonymize the specified user's account (essentially deletes their account without orphaning things)
+    archiveUser: async (_, { userId }, context: MyContext): Promise<User> => {
```
briri, line 525: Since you've added this one, can you also remove the `removeUser` resolver/schema since its doing the same thing? I think `archiveUser` is clearer

briri (changes requested):
This is looking good @jupiter007 I just had a few comments

src/models/Plan.ts
jupiter007: Good point. I made the update

src/models/Plan.ts
jupiter007: I update the comment

src/resolvers/plan.ts
jupiter007: `createdBy` was already taken, so I updated the name to be `planCreator`.

src/resolvers/plan.ts
jupiter007: Yes, good catch. Deleted.

src/resolvers/project.ts
jupiter007: They are different because they return different models (one ProjectSearchResult, and the other PlanSearchResult, and the userProjects search has different permissions. One returns plan-level data, and the other returns project-level data.

src/resolvers/user.ts
jupiter007: I believe this is already being done when we call `User.update` function: https://github.com/CDLUC3/dmptool-apollo-server/blob/development/src/models/User.ts#L491

src/resolvers/user.ts
jupiter007: Yes, makes sense. Done.

src/resolvers/user.ts
briri: ah right ok cool

briri (approved):
thanks for making those changes. Looks good

PR #775, opened 2026-06-29
Addressed bug in updateAffiliation resolver

jupiter007:
## Description

Previously, on `dev` and `stage` it appeared that `requestFeedback` functionality had failed because the user would get a `Something went wrong` error. But what I found based on the debugging I added, was that this was happening because the affiliation had no`feedbackEmails` in the database table. So, we added better messaging to the client when that happens.

Also, when I then tried to add `feedbackEmails` on the "Feedback Options" page, I was getting the same "Something went wrong" errors. This was because `updateAffiliations` was erroring on there being no `subHeaderLinks` or `ssoEmailDomains` in the `input`. We added checks for this so that it doesn't break.

- Updated `requestFeedback` with better error messaging for when `feedbackEnabled` is false or there are no `feedbackEmails`. 
- Added checks for `input.subHeaderLinks` and `input.ssoEmailDomains` in `updateAffiliations` since it was breaking that mutation when those fields were not included. 
- Removed the debugging I had previously added to investigate the `requestFeedback` resolver failing.

Fixes # ([285](https://github.com/CDLUC3/dmptool-doc/issues/285))

## Type of change

## How Has This Been Tested?
Manually tested

## Checklist:
- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/resolvers/feedback.ts
+++ b/src/resolvers/feedback.ts
@@ -180,16 +180,6 @@ export const resolvers: Resolvers = {
 
             const affiliation = await Affiliation.findByURI(reference, context, affiliationId);
 
-            if (!affiliation.feedbackEnabled) {
-              context.logger.warn({ affiliationId }, `${reference}: feedback is not enabled for affiliation`);
-              throw ForbiddenError(`Feedback is not enabled for your organization. Please enable it in your feedback settings.`);
-            }
-
-            if (affiliation.feedbackEmails.length === 0) {
-              context.logger.warn({ affiliationId }, `${reference}: no feedback emails configured`);
-              throw ForbiddenError(`No feedback recipients are configured for your organization. Please add at least one email in your feedback settings.`);
-            }
-
             const planURL = `/projects/${project.id}/dmp/${planId}`;
             const planOwnerName = [context.token.givenName, context.token.surname].filter(Boolean).join(' ');
             const planTitle = plan.title || 'Untitled Plan';
--- a/src/resolvers/feedback.ts
+++ b/src/resolvers/feedback.ts
@@ -153,22 +153,17 @@ export const resolvers: Resolvers = {
 
       try {
         if (isAuthorized(context.token)) {
-          context.logger.info({ planId, userId: context.token.id, affiliationId: context.token.affiliationId }, `${reference}: authorized, starting`);
-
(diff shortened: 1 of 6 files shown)


src/resolvers/feedback.ts
```diff
@@ -153,52 +153,46 @@ export const resolvers: Resolvers = {
...
-            );
+
+            if (!affiliation.feedbackEnabled) {
+              context.logger.warn({ affiliationId }, `${reference}: feedback is not enabled for affiliation`);
+              throw ForbiddenError(`Feedback is not enabled for your organization. Please enable it in your feedback settings.`);
+            }
 
             if (affiliation.feedbackEmails.length === 0) {
               context.logger.warn({ affiliationId }, `${reference}: no feedback emails configured`);
+              throw ForbiddenError(`No feedback recipients are configured for your organization. Please add at least one email in your feedback settings.`);
```
briri: I'm not sure this is the right place for this, at least not a message like this since a researcher wouldn't understand or be able to fix themselves.

We shouldn't ever have this scenario anyway. The admin should be required to provide at least one email when enabling feedback.

If this is due to migrated data not having feedback emails, we can run a script to fix that.

If there are no feedback emails defined, we probably should default to send the email to the affiliation's contactEmail. Maybe a good question for Becky.

briri (changes requested):
I think we need to maybe run a script to do some cleanup of migrated data if these feedback emails are missing. We should also probably just default to the `contactEmail` defined on the affiliation if none are defined.

src/resolvers/feedback.ts
jupiter007: @briri good point about the researcher not understanding the message. Maybe we should disable "request feedback" when the "feedbackEnabled" is false or there are no "feedbackEmails". That way the user won't be put in that situation. I'll check with Becky

briri: Yeah, I think that would make sense

jupiter007: Thanks @briri. I removed the error messages from `requestFeedback` and, instead, updated `nextJS` app to handle disabling of `request feedback` link when feedback is disabled: https://github.com/CDLUC3/dmptool-ui/pull/1325

briri (approved):
Thanks for making those changes @jupiter007

Issue #286, opened 2025-05-22
BE support for selecting preferred repositories, standards, etc


bofstein:
Build schema and resolves for allowing the template write to select preferred list of repositories, licensees, metadata standards)

Similarly we want the template write to select the options for Output Type (which will be the only ones offered, not just preferred)

PR #776, opened 2026-06-29
Bug/plan owner

briri:
Fixes [#286](https://github.com/CDLUC3/dmptool-doc/issues/286)

- Added a new directory to store helpful SQL scripts to repair data and assist with debugging
- Added script to ensure that every plan has a primary owner
- Added a script to repair research output question's JSON content

The scripts have already been run in dev and stage


src/scripts/sql/plan-fix-missing-planOwners.sql
```diff
@@ -0,0 +1,76 @@
...
+      ) AS primary_check ON primary_check.id = projects.id
+  ) AS missing_primary
+  WHERE missing_primary.first_owner IS NOT NULL
+);
+
+-- Add the designated user as the primary owner for all projects that did not have at least one owner
+-- Basically everything else that has no PRIMARY owner.
+SET @user_id := <a valid user id>;
+INSERT INTO projectCollaborators (projectId, email, invitedById, userId, accessLevel, created, createdById, modified, modifiedById)
+(SELECT p.id, super.email, super.id, super.id, 'PRIMARY', CURDATE(), super.id, CURDATE(), super.id
```
jupiter007: L47: Do we want to use `NOW()` instead of `CURDATE()` because of the TIMESTAMP types for dates?

jupiter007 (approved):
Looks good. Thanks Brian. Good idea to save some scripts for future use.

I just had one question about using `CURDATE()` instead of `NOW()`

src/scripts/sql/plan-fix-missing-planOwners.sql
briri: 👍🏻 yeah, I will update. If you use `CURDATE` it works, it just sets the time to `00:00:00`. We should use the actual time though

PR #779, opened 2026-07-07
Updated VersionedGuidance and Guidance models to make tagId be optional

jupiter007:
## Description

Trying to publish a guidanceGroup on `dev` was consistently failing, although I could not reproduce it locally.

After looking at the logs, the issue appeared to be that none of the `guidance` records on `dev` have `tagId` values, and it was required in the `Guidance` and `VersionedGuidance` models, even though the `guidance` table has it as optional (can be `NULL`).

- Updated `Guidance` and `VersionedGuidance` classes to make `tagId` optional, since it was causing errors on `dev` where `guidance` records had no `tagId`. This is consistent with the `guidance` table schema which allows the `tagId` field to be `NULL` 

Fixes # ([54](https://github.com/CDLUC3/dmptool-doc/issues/54))

## Type of change

## How Has This Been Tested?
Tests passing

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,7 @@
 - added data-migration to fix question JSON so that `"selected": 0` is now `"selected": false` (and `1` -> `true`).
 
 ### Updated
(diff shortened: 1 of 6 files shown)


briri (approved):
LGTM

PR #782, opened 2026-07-13
Added EZID config and datasource for registering dmpIds with EZID

jupiter007:
## Description

- Added files for EZID creation: `config/ezidConfig.ts` and `datasources/EZIDAPI.ts`, and updated `Plan.publish` to call on the new `registerIdentifier`

Once I published a plan, I could find it on `ezid-stg.cdlib.org`

Fixes # ([32](https://github.com/CDLUC3/dmsp_api_prototype/issues/32))

## Type of change

## How Has This Been Tested?
Manually tested with EZID stage, and added unit tests.

Once a plan was published, I could find the DOI on `ezid-stg.cdlib.org`. Here is an example of one that was registered: https://ezid-stg.cdlib.org/id/doi:10.48321/D15B418AFE

## Checklist:


src/datasources/EZIDAPI.ts
```diff
@@ -0,0 +1,84 @@
...
+  override willSendRequest(_path: string, request: AugmentedRequest) {
+    const credentials = Buffer.from(
+      `${EZIDConfig.username}:${EZIDConfig.password}`
+    ).toString('base64');
+    request.headers['authorization'] = `Basic ${credentials}`;
+    request.headers['content-type'] = 'text/plain; charset=UTF-8';
+  }
+
+  // Create an identifier with EZID via PUT /id/{identifier}.
+  // The identifier must be in EZID format, e.g. "doi:10.48321/D1xxxxxxxx".
```
briri: Although our DOI shoulder is not a secret per se, can you change this comment to `doi:11.22222/A1B2C3d4`

src/models/Plan.ts
```diff
@@ -757,6 +757,47 @@ export class Plan extends MySqlModel {
...
+          // Convert the stored URL form (https://doi.org/10.x/y) to EZID form (doi:10.x/y)
+          const ezidIdentifier = this.dmpId.replace(
+            generalConfig.dmpIdBaseURL,
+            'doi:'
+          );
+          const doiSuffix = ezidIdentifier.replace('doi:', '');
+          const domain = generalConfig.domain.startsWith('http')
+            ? generalConfig.domain
+            : `https://${generalConfig.domain}`;
+          const creatorName = [context.token.givenName, context.token.surName]
```
briri: according to the EZID documentation we should use `Surname Givenname`

src/models/Plan.ts
```diff
@@ -757,6 +757,47 @@ export class Plan extends MySqlModel {
...
+            .filter(Boolean)
+            .join(' ') || 'Unknown';
+          const metadata: Record<string, string> = {
+            '_profile': 'datacite',
+            '_target': `${domain}/dmps/${doiSuffix}`,
```
briri, line 790: this isn't going to be enough data. You're missing the user's ORCID, affiliation information, funding info, additional contributors (members not collaborators), etc.

Its crucial that we include any and all ORCIDs and RORs that we have available.

The resourcetype should be `OutputManagementPlan/Data Management Plan`.

Refer to the Lambda code referenced in the ticket and also see this example from the current production system: https://ezid.cdlib.org/manage/display_xml/doi:10.48321/D106377D3C

I don't see a lot of documentation beyond what you have here for ANVL, so you may just want to try things like `creator.nameIdentifier` to see if they work or not, and see any errors (or of course, ask Jing)

briri (changes requested):
thanks @jupiter007 we need to make sure we're including all of the data that we currently publish in production

src/models/Plan.ts
jupiter007: Ah ok. I thought for this ticket we were just doing the minimum required data. I will check to see what other data I can include. 

Thanks for the review Brian.

src/services/dataciteXMLService.ts
```diff
@@ -0,0 +1,369 @@
...
+// not a full locale — unlike xml:lang attributes, which accept locale tags
+// like "en-US" directly. Plan.languageId is stored as a locale (e.g. "en-US"),
+// so this strips it down for use in <language>.
+function twoCharLanguage(languageId?: string): string {
+  if (!languageId) return 'en';
+  return languageId.split('-')[0].toLowerCase() || 'en';
+}
+
+// Strict ORCID format: 4 groups of 4 digits, last group's final char may be X.
+const ORCID_FORMAT = /^\d{4}-\d{4}-\d{4}-\d{3}[\dX]$/;
```
briri: There are already a few functions in `utils/herper.ts` for dealing with ORCIDs that you can probably use

briri (approved):
This looks good @jupiter007. I just had one additional comment.

src/services/dataciteXMLService.ts
jupiter007: Thanks Brian.  I imported `ORCID_REGEX` from the `helpers.ts` file and replaced `ORCID_FORMAT` with it. Some of the other functions from that helper file were a bit different so I didn't implement them in fear that it would break some of the other code.

Issue #261, opened 2025-04-28
Sections all have "Copy Of" at the start that shouldn't be there


bofstein:
We should not say "copy of" on the Section names.

Also please confirm if these sections are truly blank, or if the Questions aren't being copied

![Image](https://github.com/user-attachments/assets/cf7365f2-ed63-4448-a5af-8aa4b5cd85bb)

Issue #241, opened 2025-03-25
Display Logic backend support


bofstein:
This is a ticket for the backend work we need for display logic.  At launch this only needs to support conditions based on a single past question.  But ideally it is designed in a way to make it extensible in the future to triggering off of multiple questions, or off of metadata (e.g. display only if Funder is X or if there are more than 5 Project Contrbiutors)

jupiter007:
Display Logic Architecture

A [Question](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) now carries two fields — [displayLogicAction](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (SHOW_QUESTION / HIDE_QUESTION) and [displayLogicMatchType](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (ANY / ALL) — that control how its condition groups are evaluated.

A [QuestionConditionGroup](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) represents one "qualifying trigger": it points to a triggerQuestion whose answer is being tested, and owns one or more [QuestionCondition](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) rows that define the individual checks (e.g. answer equals "Yes", answer contains "Data").

The [displayLogicMatchType](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) on the question determines whether any group must pass or all groups must pass to trigger the action.

Question
├── displayLogicAction      (SHOW_QUESTION | HIDE_QUESTION)
├── displayLogicMatchType   (ANY | ALL)
└── QuestionConditionGroup[]           -- one per trigger question
    ├── triggerQuestionId              -- which prior question to check
    └── QuestionCondition[]            -- the individual checks (OR'd together)
        ├── conditionType  (EQUAL | NOT_EQUAL | INCLUDES | …)
        └── conditionMatch             -- the value to match against

[...]

PR #794, opened 2026-07-23
Support for Adding/Updating/Deleting an entire Plan at once

briri:
## Description

Support for Adding, Updating and Deleting an entire Plan in once mutation. Part of [#261](https://github.com/CDLUC3/dmptool-doc/issues/261) and [#265](https://github.com/CDLUC3/dmptool-doc/issues/265)

Also provides MySQL transaction support [#241](https://github.com/CDLUC3/dmptool-doc/issues/241)

**Database**
- Fixed issue with JSON of default questions and answers in the local data migration file
- Added new helper SQL script to delete all existing research output questions and answers
- Added new `TransactionClient` class to the `datasources/mysql.ts` file to allow for the use of MySQL transactions

**Schema**
- Added a `AddAnswerInput`, `AddProjectInput`, `EntirePlanProjectFragment`, `EntirePlanMemberFragment`, `EntirePlanFundingFragment`, `EntirePlanAnswerFragment`, `AddEntirePlanInput`, `UpdateEntirePlanInput` to schema 

**Resolvers and Services**
- Added new `addEntirePlan`, `updateEntirePlan` and `removeEntirePlan` schema and resolvers to allow a caller to include all of the Project, Plan, Member, Funding, Answer and RelatedWork information in one mutation (to support REST API functionality and future integrations)
- Added new `entirePlanService` to support the above resolvers

**Models**
- Renamed old `Funding.findByProjectFundingId` to `Funding.findByPlanAndProjectFundingId` 
- Added a new `Funding.findByProjectFundingId`
- Added a new `errorsToString` function to the `MySQLModel`
- Added a new `findByOwnerAndTitle` to the `Plan` model
- Removed unused `valueIsDate` function from the `MySQLModel` 
- Removed old `processResult` functions from the `Answer` and `Question` models. These functions were added to help add the `commonStandardId` to existing JSON records. It proved to be inadequate so we eneded up just deleting old data

## Type of change

## How Has This Been Tested?

Added a lot of new unit tests

## Checklist:


src/__mocks__/context.ts
```diff
@@ -140,11 +167,30 @@ export const mockSuperAdminToken = async (): Promise<JWTAccessToken> => {
...
-  ezidAPIDataSource: new EZIDAPI({ cache: null }),
-  sqlDataSource: mockedMysqlInstance,
+// Lazy create mockDataSources to allow mocks to be set up first
+interface MockDataSources {
+  dmphubAPIDataSource: DMPHubAPI;
+  ezidAPIDataSource: EZIDAPI;
+  sqlDataSource: jest.Mocked<MySQLConnection>;
 }
+let cachedDataSources: MockDataSources | null = null;
+export const getMockDataSources = () => {
```
briri, line 177: Switched to this cached/singleton approach to hopefully help speed up tests

src/datasources/mysql.ts
```diff
@@ -1,18 +1,32 @@
...
+ * A database error
+ */
 export class DatabaseError extends Error {
-  constructor(message: string, public readonly originalError?: Error) {
+  public readonly code?: string;
+
+  constructor(message: string, public readonly originalError?: unknown) {
     super(message);
     this.name = 'DatabaseError';
+    if (originalError && typeof originalError === 'object' && 'code' in originalError) {
```
briri, line 26: Codepilot recommended this refactor. I guess it gives us some type safety and ensures the original error code isn't lost

src/datasources/mysql.ts
```diff
@@ -37,9 +68,6 @@ export class MySQLConnection implements DatabaseConnection {
         ...POOL_CONFIG
       });
 
-      // Add initialization check
-      this.initPromise = this.validateConnection();
```
briri, line 41: This was uneccessary. Using our `validateConnection` function instead

src/models/__tests__/Affiliation.spec.ts
```diff
@@ -125,7 +125,7 @@ describe('findById', () => {
     const result = await Affiliation.findById('Test', context, affiliation.id);
     const expectedSql = 'SELECT * FROM affiliations WHERE id = ?';
     expect(localQuery).toHaveBeenCalledTimes(1);
-    expect(localQuery).toHaveBeenLastCalledWith(context, expectedSql, [id.toString()], 'Test')
+    expect(localQuery).toHaveBeenLastCalledWith(context, expectedSql, [id.toString()], 'Test');
```
briri, line 128: Sorry for all the noise with these tests. 

I had originally set things up so you could pass a `TransactionClient` arround to things like `Plan.findById` and `plan.update` but then ended up refactoring to use the `withTransaction` function built into the data source. 

The changes you see here are mostly just the additions or missing semicolons, values for fields that are defaulted, etc. You can ignore the majority of the test files

src/models/Answer.ts
```diff
@@ -145,62 +144,8 @@ export class Answer extends MySqlModel {
     }
   }
 
-  // Process the raw question data that was returned from the database
```
briri, line 148: This is no longer necessary now that we're deleting the old stale data. We will make sure we correctly migrate in the future

src/models/Funding.ts
```diff
@@ -171,14 +171,14 @@ export class PlanFunding extends MySqlModel {
 
     // First make sure the record is valid
     if (await this.isValid()) {
-      const current = await PlanFunding.findByProjectFundingId(reference, context, this.planId, this.projectFundingId);
+      const current = await PlanFunding.findByPlanAndProjectFundingId(reference, context, this.planId, this.projectFundingId);
```
briri, line 174: Renamed to be more accurate since I need to add an actual `findByProjectFundingId`

src/models/MySqlModel.ts
```diff
@@ -50,7 +52,21 @@ export class MySqlModel {
     this.errors[property] = error;
   }
 
-  // Indicates whether or not the standard fields on the record are valid
+  /**
+   * Helper function to concatenate all errors into a single string
+   *
+   * @returns a concatenated string
+   */
+  errorsToString(): string {
```
briri, line 58: Added this so that I could pull all of the errors for associated objects like `ProjectMember` together so that they could then be attached to a single `plan.errors['members']` message and returned to the caller.

src/models/MySqlModel.ts
```diff
@@ -66,12 +82,6 @@ export class MySqlModel {
     return Object.keys(this.errors).length === 0;
   }
 
-  // Check whether or not the value is a Date
-  static valueIsDate(val: string): boolean {
```
briri, line 70: Found that this wasn't being used and we have a similar one already in the utils file

src/datasources/mysql.ts
```diff
@@ -109,6 +134,59 @@ export class MySQLConnection implements DatabaseConnection {
...
+    const txClient = new TransactionClient(connection);
+    let result: Awaited<T>;
+
+    try {
+      context.logger.debug('Starting database transaction');
+      await txClient.begin();
+      // Store current transaction in context for nested calls if needed
+      context.activeTransaction = txClient;
+
+      result = await action(txClient);
```
jupiter007, line 159: L160: This line assigns a value to `result` only if the promise is resolved. But in the case where something like `addEntirePlan` throws an error like `BadRequestError`, it goes directly to its `catch` block, and looks like that would go directly to `withTransaction`'s `catch` block without assigning a value to `result`. So, it seems like the error would be swallowed.

src/services/entirePlanService.ts
```diff
@@ -0,0 +1,1055 @@
...
+          reference,
+          context,
+          project.id,
+          m.affiliationId
+        );
+        return funding || new ProjectFunding(m);
+      },
+      getPlanObjProjectObjId: (pm: PlanFunding): number => pm.projectFundingId,
+      isUsedByOtherPlans: async (id: number): Promise<boolean> => {
+        return (await PlanFunding.findByProjectFundingId(reference, context, id)).length > 0;
```
jupiter007, line 480: L475: Did you mean to use the new `findByPlanAndProjectFundingId` here?

jupiter007 (approved):
@briri so glad you added the mysql transaction wrapper. It looks great. Just had one concern about what is actually returned by the `withTransaction` `catch` block in the cases where a promise never resolves.

src/datasources/mysql.ts
briri: Ah thanks and good catch @jupiter007. I've updated it to always rethrow the error. I handle the determinination of whether to return the plan with its errors or throw an error up to the resolver that's using `withTransaction`

src/services/entirePlanService.ts
briri: Nope. That particular function will check to make sure its not related to another plan, so we just want to load all of the ones associated with the project

PR #796, opened 2026-07-27
Add password reset token logic

jupiter007:
## Description
I thought about different ways to handle password resets and decided it would be good to create a new `passwordResetTokens` table, and have a separate model and resolver for it, since it is it's own thing.

In order to invalidate all other tokens on a password reset, I decided to use `tokenVersion` system in the redis cache. That way, we can check, without a database call, whether a user's log in is still valid on a password reset.

- Added `getUserTokenVersion` and `bumpUserTokenVersion` to `tokenService` so that we can save the `tokenVersion` in the `JWT` payload, and validate it against the current version in `isRevokedCallback` on `/graphql` requests. 
- Added new `PasswordResetToken` model, `passwordReset` resolver, `passwordReset` schema, and separate `passwordResetTokens` table to store the `resetPasswordToken` and `resetPasswordExpiresAt`
- Added `passwordChangedAt` field to the `users` table, and `User` model to record when password last changed
- Updated `emailService` with a `sendResetPasswordEmail` to send an email with the `change password link`

Fixes # ([133](https://github.com/CDLUC3/dmptool-apollo-server/issues/133))

## Type of change

## How Has This Been Tested?
Manually tested in Apollo Explorer and with added unit tests

## Checklist:


src/models/PasswordResetToken.ts
```diff
@@ -0,0 +1,114 @@
+import { MySqlModel } from './MySqlModel';
+import { MyContext } from '../context';
+import { getCurrentDate } from '../utils/helpers';
+import { prepareObjectForLogs } from '../logger';
+
+export class PasswordResetToken extends MySqlModel {
+  public userId: number;
+  public resetPasswordToken?: string;
+  public resetPasswordExpiresAt?: string;
+  public usedAt?: string;
```
briri, line 13: Why would we keep these? I think simply deleting them after they've been used (or invalidated) would be better. 
I cannot see a scenario where we'd need to reference them again.

src/models/PasswordResetToken.ts
```diff
@@ -0,0 +1,114 @@
...
+    const newToken = new PasswordResetToken({ userId, resetPasswordToken: resetToken, resetPasswordExpiresAt: expiresAt });
+    const saved = await newToken.create(context);
+
+    context.logger.debug(prepareObjectForLogs({ userId, savedId: saved?.id }), reference);
+    return saved?.id ? saved : null;
+  }
+
+  // Find the still-valid token by the passwordResetToken record ID
+  static async findById(reference: string, context: MyContext, id: number): Promise<PasswordResetToken | null> {
+    const sql = 'SELECT * FROM passwordResetTokens WHERE id = ?';
```
briri: This may be fine as is, but the comment says "still-valid". If you want to restrict it, you should add an expiry check like you have in the one below

src/models/User.ts
```diff
@@ -362,6 +365,25 @@ export class User extends MySqlModel {
...
+    if (this.id) {
+      this.password = newPassword;
+      if (!this.validatePassword()) {
+        return false;
+      }
+      this.password = await this.hashPassword(newPassword);
+      this.passwordChangedAt = getCurrentDate();
+
+      if (await User.update(context, this.tableName, this, 'User.setPassword')) {
+        await bumpUserTokenVersion(context.cache, this.id);
```
briri, line 379: Would prefer to see this in the resolver. We handle all other token activity in the services and resolvers.

src/resolvers/passwordReset.ts
```diff
@@ -0,0 +1,92 @@
...
+        return false;
+      }
+    },
+  },
+  Mutation: {
+    // Send a password reset email to the user with a valid account
+    sendPasswordResetEmail: async (_, { email }, context: MyContext): Promise<boolean> => {
+      const reference = 'sendPasswordResetEmail resolver';
+      try {
+        const userEmail = await UserEmail.findByEmail(reference, context, email);
```
briri: Do we want to limit this so that it will only work with a user's primary email? 
1. Its the only one the user logs in with
2. Might be safer than using an alternate email in the event that the user had a typo when they added a secondary email and "confirm your email" message went to a bad actor who then decides to try and reset the password and hijack the account.

src/resolvers/passwordReset.ts
```diff
@@ -0,0 +1,92 @@
...
+        const user = userEmail[0] ? await User.findById(reference, context, userEmail[0].userId) : null;
+
+        if (user && user.active && !user.locked) {
+          const rawToken = crypto.randomBytes(32).toString('hex');
```
briri: I think this token generation belongs within the model.

src/resolvers/passwordReset.ts
```diff
@@ -0,0 +1,92 @@
...
+        if (!user || !user.active || user.locked) {
+          throw ForbiddenError();
+        }
+
+        const passwordSet = await user.setPassword(context, newPassword);
+        if (!passwordSet) {
+          throw InternalServerError();
+        }
+
+        const marked = await resetTokenRecord.markUsed(context);
```
briri, line 83: I think we can just go ahead and delete the tokens.

src/resolvers/user.ts
```diff
@@ -431,31 +435,6 @@ export const resolvers: Resolvers = {
       }
     },
 
-    // Change the current user's password
```
briri, line 434: Why delete this resolver? This is intended for the update password page in the account profile section.

Its also in the schema so I'm surprised it didn't raise errors

src/models/User.ts
briri: I'm also not sure we need to introduce the tokenVersion. Why not just invalidate the existing token by calling `revokeAccessToken` like we do in the signout controller?

Requiring people to re-authenticate after a password reset is a pretty common pattern.

briri (changes requested):
I like the approach with the DB table @jupiter007.

I wonder though if we need the token version logic at all and wether we need to record the `usedAt` timestamp and retain these tokens.

briri (approved):
Thanks for making those changes @jupiter007.

I didn't see any of your responses to my comments which is weird. Anyway, I was thinking about it this morning and it does make sense to hang onto the expired/used tokens in the DB table so we can identify patterns of abuse in the future. 

I'll add a ticket to add a cleanup job so we can purge records after a period of time (e.g. 30 days). We're going to want a job like that to clean up a few tables I think

PR #809, opened 2026-08-07
Feature/508/js add question conditions table and resolvers

jupiter007:
## Description

Display Logic Architecture

A [Question](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) now carries two fields — [displayLogicAction](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (SHOW_QUESTION / HIDE_QUESTION) and [displayLogicMatchType](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (ANY / ALL) — that control how its condition groups are evaluated.

A [QuestionConditionGroup](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) represents one "qualifying trigger": it points to a triggerQuestion whose answer is being tested, and owns one or more [QuestionCondition](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) rows that define the individual checks (e.g. answer equals "Yes", answer contains "Data").

The [displayLogicMatchType](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) on the question determines whether any group must pass or all groups must pass to trigger the action.

Question
|── displayLogicAction      (SHOW_QUESTION | HIDE_QUESTION)
|── displayLogicMatchType   (ANY | ALL)
└── QuestionConditionGroup[]           -- one per trigger question
    |── triggerQuestionId              -- which prior question to check
    └── QuestionCondition[]            -- the individual checks (OR/AND together)
        |── conditionType  (EQUAL | NOT_EQUAL | INCLUDES | …)
        └── conditionMatch             -- the value to match against

When a template is published, the live groups/conditions (which get wiped on every saveQuestionDisplayLogic call) are snapshotted into [VersionedQuestionConditionGroup](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [VersionedQuestionCondition](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), linked to the [VersionedQuestion](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) rather than back to the transient live records.

## Changes
- Added `questionConditionGroups` and `versionedQuestionConditionGroups` tables via data migration scripts
- Added new models `QuestionConditionGroup` and `QuestionConditionGroups` models for question display logic
- Updated `versionedQuestion` resolver to use `findByVersionedQuestionConditionGroupId` in place of `findByVersionedQuestionId`
- Updated `publishedConditionGroupsForQuestion` in `versionedQuestionCondition` resolver and added `conditions` chained resolver
- Updated `question`, `questionCondition` and `versionedQuestionCondition` schemas
- Updated `Question` model to include `displayLogicAction` and `displayLogicMatchType`. 
- Updated `QuestionCondition` model to only include `conditionType`, `conditionMatch` and `groupId`, and replaced `findByQuestionId` with `findByGroupId`. 
- Updated `VersionedQuestionCondition` model to use `versionedQuestionConditionGroupId instead of `versionedQuestionId`
- Altered `questions`, `questionConditions`, `versionedQuestions` and `versionedQuestionConditions` tables via data migration scripts
- Updated `questionService.ts` with `generateQuestionConditionGroupVersion`
- Fixed an issue in `MySqlModel` where the `query` function was returning a tuple with fields along with the response 

Fixes # ([508](https://github.com/CDLUC3/dmptool-apollo-server/issues/508))([509](https://github.com/CDLUC3/dmptool-apollo-server/issues/509))

## Type of change

## How Has This Been Tested?
Manual testing with frontend changes, and also with new/updated unit tests

## Checklist:
* There are frontend changes that consume this but nothing preventing merging of these changes


src/models/QuestionCondition.ts
```diff
@@ -1,44 +1,34 @@
 import { MyContext } from "../context";
 import { MySqlModel } from "./MySqlModel";
+import {
```
jupiter007: Each condition added to a qualifying options questions, which occurs before the question being edited, will be added to `questionConditions` table. The `questionConditions` are always tied to a `questionConditionGroup`, since the qualifying options questions can have multiple conditions applied to it.

briri (approved):
I have some reservations about placing the `displayLogicAction` and `displayLogicMatchType` on the question itself for 2 reasons.

1: I feel that having them on the question could be very confusing. For example the default is "SHOW_QUESTION" which (thinking from an external system's standpoint) would leave me to believe that I should show this question by default when the reverse is actually what we want (meaning we only want it to show if some set of triggers are true)

2: I think this could trap us into a scenario in the future when we start to allow more complex trigger types like if a text area contains a certain word/phrase, etc. Or if we end up having a hierarchy of logic like "hide this question if the user answered "Apple" on question 4" but also has a higher priority one like "always show this question if the user answers "Yes" to question 1.

Since I can't think of a good scenario though, lets go ahead and merge it for the demo. I'd like to think more about potential scenarios and future expansion of the logic

PR #811, opened 2026-08-11
Added a new plan resolver and some functions in planService for Landing Page

jupiter007:
## Description

- Added `publicPlanbyDMPId` resolver, to query only public plans for the landing page
- Added `getPlanVersions` and `extractPlanOutputs` for the `publicPlanByDMPId` resolver

Fixes # ([293](https://github.com/CDLUC3/dmptool-ui/issues/293))

## Type of change

## How Has This Been Tested?
Manually tested with the corresponding Frontend branch, https://github.com/CDLUC3/dmptool-ui/pull/1394, and added unit tests

## Checklist:


briri (changes requested):
We show a landing page ONLY for published/registered plans. This is because a DOI needs a place to resolve to. For example [https://doi.org/10.48321/D174A1D243](https://doi.org/10.48321/D174A1D243) resolves to [this landing page](https://dmphub.uc3prd.cdlib.net/dmps/10.48321/D174A1D243)

We show the landing the page regardless of the visibility setting. The only thing we currently restrict is whether or not the narrative PDF is downloadable.

You should also be able to just get the outputs from the maDMP JSON record directly. No need to construct them in the service.

briri (approved):
thanks for making that change

PR #357, opened 2025-07-28
fixed issue where apiTarget could not be null in the graphQL schema f…

briri:
## Description

Fixed a bug in the popular funders query that was requiring `apiTarget` to be non-null

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Tested in Apollo explorer

## Checklist:
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I updated the CHANGELOG.md and added documentation if necessary
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #814, opened 2026-08-19
bug fixes for create entirePlan

briri:
## Description

Part of fix [#357](https://github.com/CDLUC3/dmptool-doc/issues/357)

- Bumped `@dmptool/utils` to 2.1.7
- Fixed bug with SQL query in `ProjectMember.findByProjectAndNameOrORCIDOrEmail`
- Added missing call to generate the maDMP JSON after successful creation or update of the Plan.
- Fixed some minor issues in the `entirePlanService`

## Type of change

## How Has This Been Tested?

All tests are passing and tested manually through the REST API app

## Checklist:

### DIFF
--- a/src/services/entirePlanService.ts
+++ b/src/services/entirePlanService.ts
@@ -237,8 +237,8 @@ export const processMemberAssociations = async(
     {
       // Define all the fetch functions that will give us the information we need
       // to determine whether an association should be added, updated or removed
-      fetchPlanObjs: async (id: number): Promise<PlanMember[]> => {
-        return await PlanMember.findByPlanId(reference, context, id);
+      fetchPlanObjs: (id: number): Promise<PlanMember[]> => {
+        return PlanMember.findByPlanId(reference, context, id);
       },
       fetchProjectObjs: (id: number): Promise<ProjectMember[]> => {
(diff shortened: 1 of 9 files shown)


src/services/entirePlanService.ts
```diff
@@ -828,22 +831,23 @@ export const addEntirePlan = async (
...
     // 2nd: find or initialize the project
-    const project: Project | undefined = await findOrInitializeProject(reference, context, input.project);
+    let project: Project | undefined = await findOrInitializeProject(reference, context, input.project);
     if (!project) {
       context.logger.fatal(logBase, 'Unable to find or initialize a Project!');
       throw InternalServerError();
     }
 
     // 3rd: Save the project
     if (project.id) {
```
jupiter007: The only thing I would say here is using the `project` variable name for both the existing and newly saved is sort of confusing. It might be more clear if you did something like 
```
const savedProject = project.id
  ? await project.update(context, false)
  : await project.create(context);
```
and then 
```
if (savedProject.hasErrors()) {
  plan.addError('projectId', savedProject.errorsToString());
  throw BadRequestError();
}
```

jupiter007 (approved):
Looks good. Just had one small comment.

PR #815, opened 2026-08-23
Bug/339/js fix bugs for landing page

jupiter007:
## Description

Fixed versioning issues. Was able to use the new `AcceptedWork` class to get related works.

- Updated `updateAnswer` resolver to update `plan.modified` to reflect that something had changed (for versioning). Otherwise, it wasn't registering as a version change.
- Added `isPrimaryContact` to UpdateProjectMemberInput because the `isPrimaryContact` field value was being overwritten to `false` when the input didn’t include it
- Added `owner` and `relatedWorks` to `PlanVersionSnapshot` schema - PlanVersionSnapshot is the schema for what is needed for the landing page. The `maDMPToolDMPToSnapshot` transforms the data into a shape that can easily be utilized by the client.
- Updated `getPlanVersions` in `planService` to exclude the `VERSION#latest` used for the dropdown list. Its not a queryable timestamped snapshot in and of itself.
- Updated `getPlanVersionSnapshot` to fetch the `planId` and `projectId` to pass to `mapDMPToolDMPToSnapshot`
- Updated `maDMPToolDMPToSnapshot` to get `isPrimaryContact` from correct place, and added `owner` and `relatedWorks` to the response

Fixes # ([339](https://github.com/CDLUC3/dmptool-doc/issues/339))

## Type of change

## How Has This Been Tested?
Manually tested with Apollo Explorer and frontend changes, and added new unit tests

## Checklist:

## Screen recording

https://github.com/user-attachments/assets/45ee98ec-158f-45d7-8bcf-e34c5ba50979

### DIFF
--- a/src/resolvers/answer.ts
+++ b/src/resolvers/answer.ts
@@ -156,6 +156,10 @@ export const resolvers: Resolvers = {
             const updatedAnswer = await answer.update(context);
 
             if (updatedAnswer && !updatedAnswer.hasErrors()) {
+              // Update the plan's modified timestamp to reflect that something changed
+              plan.modified = new Date().toISOString();
+              await plan.update(context);
+
               // Handle OpenSearch index update and maDMP JSON versioning in Dynamo
               await handleAsyncUpdates(reference, context, plan, project);
             }
--- a/src/resolvers/plan.ts
+++ b/src/resolvers/plan.ts
@@ -18,6 +18,10 @@ import { VersionedTemplate } from "../models/VersionedTemplate";
 import { Answer } from "../models/Answer";
 import { ProjectCollaboratorAccessLevel } from "../models/Collaborator";
 import { AlternateIdentifier } from "../models/AlternateIdentifier";
+import {
+  RelatedWorkSearchResult,
+  RelatedWorkStatus
+} from "../models/RelatedWork";
 import { isNullOrUndefined, normaliseDateTime } from "../utils/helpers";
 import {
   AuthenticationError,
@@ -37,7 +41,9 @@ import {
   PaginatedPlanResults,
   PlanFeedbackStatus,
   Resolvers,
-  UpdateEntirePlanInput
+  UpdateEntirePlanInput,
+  RelatedWorksFilterOptions,
+  PlanVersionSnapshot
 } from "../types";
 import { prepareObjectForLogs } from "../logger";
 // Services
@@ -45,6 +51,7 @@ import {
   buildDataCiteXMLForPlan,
   ensureDefaultPlanContact,
   getPlanVersions,
+  getPlanVersionSnapshot,
   saveMaDMPVersion
 } from "../services/planService";
 import {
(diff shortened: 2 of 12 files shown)


src/resolvers/answer.ts
```diff
@@ -156,6 +156,10 @@ export const resolvers: Resolvers = {
             const updatedAnswer = await answer.update(context);
 
             if (updatedAnswer && !updatedAnswer.hasErrors()) {
+              // Update the plan's modified timestamp to reflect that something changed
+              plan.modified = new Date().toISOString();
+              await plan.update(context);
```
briri: Good catch!

src/resolvers/plan.ts
```diff
@@ -779,6 +785,31 @@ export const resolvers: Resolvers = {
...
+      if (primaryContactMember?.projectMemberId) {
+        const projectMember = await ProjectMember.findById(reference, context, primaryContactMember.projectMemberId);
+        if (projectMember?.affiliationId) {
+          const affiliation = await Affiliation.findByURI(reference, context, projectMember.affiliationId);
+          if (affiliation) return affiliation;
+        }
+      }
+
+      // Fall back to the plan creator's affiliation
+      if (parent?.createdById) {
```
briri: I'm not sure the primary contact and owner are always the same thing. And the creator and owner are also not always the same thing.

We added the `OWN` permission level to the `Collaborator` a few months ago. I think you want to fetch that user and then fall back to `plan.createdById` if there is not one set.

src/resolvers/plan.ts
```diff
@@ -852,6 +883,41 @@ export const resolvers: Resolvers = {
...
+    ): Promise<RelatedWorkSearchResult[]> => {
+      if (!parent?.id || !parent?.projectId) {
+        return [];
+      }
+
+      const isPubliclyVisible = parent.visibility === 'PUBLIC' && !!parent.registered;
+
+      let effectiveFilters: RelatedWorksFilterOptions = {};
+
+      if (isPubliclyVisible) {
```
briri: I cannot envision a scenario where we would want someone to see related works that are not accepted on the landing page.

The current landing page's purpose is to provide a human readable page that the DOI (DMP id) resolves to. We're allowing it for unpublished plans in the new system, but its primary purpose is to provide a public facing page that provides high level info about the plan.

briri (changes requested):
I think you want to get the Collaborator who is designated as the `OWN`er. There's also not a need to show  related works that have not been `ACCEPTED` on the landing page

src/resolvers/plan.ts
jupiter007: Thanks Brian. I believe I added that when i was testing out returning related works data before the AcceptedWork class was created, and forgot to remove it.

Thanks so much for catching that. I removed it.

src/resolvers/plan.ts
jupiter007: Very good point.

I updated to use `OWN` access level if it's available, and if not, fall back to the creator of the plan.

src/resolvers/plan.ts
briri: Ah ok. cool

briri (approved):
Thanks for making those changes @jupiter007

jupiter007:
> Thanks for making those changes @jupiter007

Thanks Brian

PR #816, opened 2026-08-24
Bug/fixes for update entire plan resolver

briri:
## Description

Part of Fix for [#359](https://github.com/CDLUC3/dmptool-doc/issues/359)

**NOTE:** You need to run the data migrations for this to one

**Database changes:**
- Add missing `relationType` column to the `relatedWorks` table
- Added cascading delete logic to the `relatedWorks` table so that those entries are dropped when a plan is deleted.

**Code changes:**
- Updated docker compose so that LocalStack log level is now `WARN`.
- Bumped versions of `@dmptool` packages to their latest versions
- Updated `RelatedWork` model so that it does not always try to strip off the protocol and domain from the related work identifier. The DMP works matching deals with DOIs, but we are allowing users to manually add related works via the UI as well as through the REST API and sometimes the entries are URLs or other unique identetifiers that are NOT DOIs. These changes continue to strip off the `https://doi.org` if it is a DOI (for backward compatibility with the dmp works project) but otherwise preserve the entire value
- Added chained resolver to `Plan` to return `acceptedWorks`
- Fixed issues with `entirePlanService` that was causing updates to fail.
- Moved logic from the `relatedWork` resolver into the `relatedWorkService` and renamed to `addAcceptedWork` so that adding a new work manually through the UI or via the `entirePlan` resolvers use the same logic.

## Type of change

## How Has This Been Tested?

All unit tests are passing, and verified that the REST API is able to update

## Checklist:


src/schemas/plan.ts
```diff
@@ -507,6 +517,9 @@ type PlanVersionSnapshotAnswer {
     funding: [EntirePlanFundingFragment!]
     "The answers to the questions in the plan's narrative"
     answers: [EntirePlanAnswerFragment!]
+
+    "Related Works associated with the plan"
+    acceptedWorks: [EntirePlanAcceptedWorkFragment!]
```
jupiter007: Does `acceptedWorks` always return an array? If so, you can add a `!` at the end

jupiter007 (approved):
Looks good!

